PR #4, opened 2026-02-22
Add per-job notify field to control Telegram forwarding

jowitte:
## Summary

- Adds optional `notify` field to job frontmatter (`true` | `false` | `"error"`)
- `true` (default): always forward output to Telegram (backward compatible)
- `false`: never forward (for silent housekeeping jobs like vault-sync)
- `"error"`: only forward on non-zero exit code

## Motivation

Recurring housekeeping jobs (e.g. vault-sync every 5 minutes) generate noise in Telegram. Prompt-based suppression is fragile — job names still get sent even when Claude produces no meaningful output. This adds declarative control at the job level.

## Changes

- `src/jobs.ts`: Added `notify` property to `Job` interface and parsing in `parseJobFile()`
- `src/commands/start.ts`: Conditional `forwardToTelegram()` based on `notify` field

## Usage

```yaml
---
schedule: "*/5 * * * *"
notify: false
---
Sync the vault...
```

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -635,7 +635,11 @@ export async function start(args: string[] = []) {
       if (cronMatches(job.schedule, now, currentSettings.timezoneOffsetMinutes)) {
         resolvePrompt(job.prompt)
           .then((prompt) => run(job.name, prompt))
-          .then((r) => forwardToTelegram(job.name, r))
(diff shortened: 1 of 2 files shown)


jowitte:
Thanks to Claude 😄

PR #19, opened 2026-03-09
feat(api): add message injection endpoint

dathtd119:
## Summary
- Adds `POST /api/inject` endpoint for external systems to inject messages into claudeclaw sessions via HTTP
- Bearer token authentication via `apiToken` field in settings.json
- Extends `json()` helper to support HTTP status codes

## Use Case
Allows n8n workflows (or any external system) to programmatically send messages to a running claudeclaw daemon, replacing the need for a separate hook API.

## Configuration
Add to `settings.json`:
```json
{
  "apiToken": "your-secret-token-here"
}
```

## API
```bash
curl -X POST http://127.0.0.1:4632/api/inject \
  -H "Authorization: Bearer your-secret-token-here" \
  -H "Content-Type: application/json" \
  -d '{"message": "Hello from external system"}'
```

Response: `{ "ok": true, "result": "...", "exitCode": 0 }`

## Files Changed
- `src/config.ts` — `apiToken` field in Settings
- `src/ui/http.ts` — `json()` now accepts optional status code
- `src/ui/auth.ts` — Bearer token validation (new)
- `src/ui/server.ts` — `POST /api/inject` handler

## Test plan
- [x] POST without token → 401 Unauthorized
- [x] POST with wrong token → 401 Unauthorized
- [x] POST with valid token, no message → 400 Bad Request
- [x] POST with valid token + message → 200 + assistant response
- [x] No apiToken in settings → 503 Service Unavailable

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -353,13 +353,13 @@ export async function start(args: string[] = []) {
   let telegramSend: ((chatId: number, text: string) => Promise<void>) | null = null;
   let telegramToken = "";
 
-  async function initTelegram(token: string) {
+  async function initTelegram(token: string, receiveEnabled = true) {
     if (token && token !== telegramToken) {
       const { startPolling, sendMessage } = await import("./telegram");
-      startPolling(debugFlag);
+      if (receiveEnabled) startPolling(debugFlag);
       telegramSend = (chatId, text) => sendMessage(token, chatId, text);
       telegramToken = token;
-      console.log(`[${ts()}] Telegram: enabled`);
+      console.log(`[${ts()}] Telegram: enabled${receiveEnabled ? "" : " (send-only)"}`);
     } else if (!token && telegramToken) {
       telegramSend = null;
       telegramToken = "";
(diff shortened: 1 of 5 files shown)


TerrysPOV:
### Code review

Found 3 issues:

1. Silent Telegram error suppression — `.catch(() => {})` swallows all delivery failures; the caller receives `{ ok: true }` even when the Telegram send fails, with no log entry

https://github.com/moazbuilds/claudeclaw/blob/90d7355ba31d96cd201d09db810fc74607a9074c/src/ui/server.ts#L164-L170

2. Non-constant-time Bearer token comparison — `header !== \`Bearer ${token}\`` uses direct string equality, which is vulnerable to timing attacks; `crypto.timingSafeEqual` is the correct tool here

https://github.com/moazbuilds/claudeclaw/blob/90d7355ba31d96cd201d09db810fc74607a9074c/src/ui/auth.ts#L5-L7

3. `apiToken` added to `Settings` interface but missing from `DEFAULT_SETTINGS` — every other `Settings` field has a corresponding default; this breaks the established invariant and will cause inconsistent behaviour in freshly generated `settings.json` files

https://github.com/moazbuilds/claudeclaw/blob/90d7355ba31d96cd201d09db810fc74607a9074c/src/config.ts#L83-L85

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

TerrysPOV:
Ported to #157 (feat/inject-api) with clean adaptation onto current master (which has added /api/sessions, /api/agents, /api/chat, and per-context timeouts since this PR was opened). Closing in favour of that PR.

PR #20, opened 2026-03-11
Add fork-session lane, /kill command, and auto-routing

1872-svg:
## Summary

- **Fork agent lane**: `/fork <prompt>` spawns a parallel Claude session that doesn't touch the main session or queue. Useful for checking progress or asking quick questions while the main agent is running a long task.
- **`/kill` command**: Terminates the currently running Claude subprocess immediately.
- **Auto-routing**: When the main agent is busy, incoming messages are automatically routed to the fork lane instead of queuing behind the current task.
- **Fork agent uses Haiku model** with a speed-focused system prompt — denies long tasks (compiling, downloads, fuzzing) and answers in 1-3 sentences.

## How it works

`runner.ts` exports `killActive()`, `isMainBusy()`, and `runFork()`. The fork agent runs with `--output-format json` in a separate subprocess, bypassing the serial queue used by the main session.

In `telegram.ts`, `handleMessage` checks `isMainBusy()` before dispatching — busy → fork, idle → main session.

## Test plan

- [ ] Send a long-running prompt, then `/fork what are you doing?` — should get a fast response without interrupting the main task
- [ ] Send a long-running prompt, then send another message mid-run — should be auto-routed to fork
- [ ] `/kill` while a task is running — should terminate and confirm
- [ ] `/kill` with nothing running — should report "No active agent running"

### DIFF
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -1,4 +1,4 @@
-import { ensureProjectClaudeMd, run, runUserMessage } from "../runner";
+import { ensureProjectClaudeMd, run, runUserMessage, runFork, killActive } from "../runner";
 import { getSettings, loadSettings } from "../config";
 import { resetSession } from "../sessions";
 import { transcribeAudioToText } from "../whisper";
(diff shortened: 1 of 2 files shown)


TerrysPOV:
Vision: PARTIAL — fork agent is lightweight but auto-routing silently switches to Haiku without user awareness, violating the "no hidden implicit behaviors" principle.

Overlapping PRs: #76 (modifies `activeProc` tracking in `runner.ts` — significant merge conflict risk), #82 (session isolation changes in `runner.ts` — moderate merge conflict risk)

### Code review

Found 3 issues:

1. **`activeProc` overwritten by fork: `/kill` targets wrong process** — `runFork` calls `runClaudeOnce`, which unconditionally sets `activeProc = proc`. If a fork spawns while the main agent is running, `activeProc` is overwritten to the fork's subprocess. A subsequent `/kill` kills the fork and leaves the main agent running as an untrackable orphan.

https://github.com/moazbuilds/claudeclaw/blob/0f1f920e7ad819f8696129b7bce25cb23a471e73/src/runner.ts#L109-L117

2. **Auto-routing silently switches model and context with no notification** — When `isMainBusy()` is true, the message routes to a Haiku fork with a completely different system prompt, with no indication to the user. The response comes from a different model with different capabilities and no main-session context.

https://github.com/moazbuilds/claudeclaw/blob/0f1f920e7ad819f8696129b7bce25cb23a471e73/src/commands/telegram.ts#L630-L636

3. **Fork auto-routing is always-on with no opt-in gate** — Busy-check routing activates for all users on upgrade with no `settings.json` flag to disable or opt in. This contradicts the established pattern (see #15, #17) where new routing behaviors must be opt-in.

https://github.com/moazbuilds/claudeclaw/blob/0f1f920e7ad819f8696129b7bce25cb23a471e73/src/commands/telegram.ts#L630-L636

---

Closing in favour of #76. The fork agent idea is interesting, but this implementation has two blockers: (1) auto-routing silently switches to Haiku with no user notification — violates the "no hidden implicit behaviors" principle; (2) the feature is always-on with no `settings.json` opt-in, against the pattern from #15/#17. If you want to resubmit: make routing opt-in via settings, notify the user when a message routes to the fork ("Main busy — routing to fork agent"), and keep `FORK_MODEL` configurable. Happy to review a v2.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #25, opened 2026-03-11
fix: protect job frontmatter from being overwritten during execution

### LINKED ISSUE #1: Cron frontmatter overwritten
Hey, running into a recurring issue where the YAML frontmatter gets wiped from job files in .claude/claudeclaw/jobs/. The file body stays intact but the frontmatter block ends up empty (---\n\n---), which causes the job to lose its schedule and stop being properly re-queued by the daemon.

This has happened more than once to the same job file. The job itself runs fine from cached state for a bit, but eventually falls out of the scheduler entirely.

My suspicion is that Claude is reading the job file during execution for context, then writing it back out somewhere in the process and stripping the frontmatter. The warning in the job prompt ("NEVER modify files in .claude/claudeclaw/jobs/") suggests this might be a known failure mode.

Love this project!

Aleschka:
## Summary

- Adds `snapshotJobFrontmatter()` to `jobs.ts` — takes a snapshot of the job file's YAML frontmatter before execution and returns a restore function
- Integrates snapshot/restore into the job execution loop in `start.ts` — if Claude modifies or strips the frontmatter during a run, it is automatically restored
- Handles edge cases: file deleted, frontmatter completely mangled, or body content changed by Claude (preserves updated body, restores only frontmatter)
- Logs a message when restoration occurs for debugging

Fixes #1

## Test plan

- [ ] Create a job with `schedule` and `recurring` frontmatter
- [ ] Run the job and verify frontmatter remains intact
- [ ] Manually strip frontmatter from a job file mid-run to simulate the bug, verify it gets restored
- [ ] Verify one-shot jobs still clear their schedule correctly after execution

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -4,7 +4,7 @@ import { fileURLToPath } from "url";
 import { run, runUserMessage, bootstrap, ensureProjectClaudeMd, loadHeartbeatPromptTemplate } from "../runner";
 import { writeState, type StateData } from "../statusline";
 import { cronMatches, nextCronMatch } from "../cron";
-import { clearJobSchedule, loadJobs } from "../jobs";
+import { clearJobSchedule, loadJobs, snapshotJobFrontmatter } from "../jobs";
 import { writePidFile, cleanupPidFile, checkExistingDaemon } from "../pid";
 import { initConfig, loadSettings, reloadSettings, resolvePrompt, type HeartbeatConfig, type Settings } from "../config";
 import { getDayAndMinuteAtOffset } from "../timezone";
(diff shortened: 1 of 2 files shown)


TerrysPOV (changes requested):
This change fixes one failure mode, but I found two blocking correctness issues that should be addressed before merge:

  1. snapshotJobFrontmatter() restores the entire original frontmatter whenever it differs from the post-run file. That means any legitimate concurrent edit to schedule, notify, or recurring while the job is running gets silently reverted at the end of the run.
  2. If the file is no longer parseable as frontmatter, the restore path writes back originalContent wholesale, which drops any body changes made during the run. That conflicts with the intended “restore only frontmatter, preserve updated body” behavior.

  To get this mergeable, I think the restore logic needs to be made much narrower. It should only repair the specific corruption pattern you’re trying to guard against, not treat any frontmatter difference as damage. In practice, that
  likely means:

  - preserve legitimate concurrent edits to frontmatter fields like schedule, notify, and recurring
  - avoid restoring originalContent wholesale when the file is malformed if the current body can still be recovered
  - run the frontmatter repair from cleanup/finalization so it also executes when the Claude run fails, not only on the success path
  - add tests that lock those behaviors down before merge

  Please also add automated coverage for src/jobs.ts or this scheduler path, especially for:

  - concurrent user edits to frontmatter during a run
  - malformed or missing closing --- with body changes preserved
  - one-shot jobs still clearing schedule correctly after restore

  I also couldn’t verify a clean automated check run locally as-is: bunx tsc --noEmit currently fails on existing src/config.ts signature errors unrelated to this PR. Please confirm whether you ran the test plan listed in the PR
  description and whether all applicable automated checks and manual validation steps passed locally.

TerrysPOV:
Ported to #159 (fix/job-frontmatter-session-recovery), combined with #26 since both address the same job-lifecycle reliability surface. Closing in favour of that PR.

PR #26, opened 2026-03-11
fix: prevent fallback model from corrupting resumed session

### LINKED ISSUE #18: Fallback model corrupts resumed session via thinking block signature mismatch
### Summary

When the primary model hits a rate limit and claudeclaw falls back to an alternative API provider (via `ANTHROPIC_BASE_URL`), the fallback response is saved into the resumed session history. Subsequent calls using the primary provider then fail permanently with:

```
API Error: 400 {"type":"error","error":{"type":"invalid_request_error",
"message":"messages.N.content.0: Invalid \`signature\` in \`thinking\` block"}}
```

The session becomes **permanently corrupted** — every `--resume` replays the invalid thinking block, and the API rejects it every time.

### Steps to Reproduce

1. Configure claudeclaw with a primary model and a fallback that uses a different `ANTHROPIC_BASE_URL` (e.g. a third-party proxy)
2. Start the daemon and let it accumulate a session via `--resume`
3. Wait for (or trigger) a rate limit on the primary model
4. The fallback fires and succeeds — response is saved into the session
5. On the next heartbeat/telegram/job, the primary model is used again with `--resume`
6. **Every call from this point forward fails** with `Invalid signature in thinking block`

### Root Cause

Extended thinking blocks contain **cryptographic signatures** that are provider-specific. When the fallback provider responds through a different API endpoint, its thinking block signatures are incompatible with the primary provider's API. Once that response is persisted in the session history, the session is permanently broken for the primary provider.

The fallback logic in `runner.ts` (lines 262-272) retries with a different provider on the **same session args** (including `--resume`), which mixes incompatible thinking block signatures in one conversation history.

### Secondary Issue: "Unknown" Error in Telegram/Discord Forwarding

`forwardToTelegram` (and `forwardToDiscord`) display `"Unknown"` instead of the actual error:

```typescript
: `error (exit ${result.exitCode}): ${result.stderr || "Unknown"}`
```

The `claude` CLI puts API errors in **stdout**, not stderr. So `result.stderr` is empty and users see `"error (exit 1): Unknown"` with no diagnostic info.

### Current Workaround

Delete the session file to force a fresh session:
```bash
rm .claude/claudeclaw/session.json
```

This works until the next rate limit triggers the same corruption cycle.

### Proposed Fixes

**Option A: Separate session for fallback (recommended)**

When falling back, don't pass `--resume`. Instead, run a standalone one-shot call for the fallback provider. This keeps the primary session history clean and avoids signature mixing entirely.

```typescript
// On fallback: strip --resume from args, run as one-shot
const fallbackArgs = args.filter(a => a \!== "--resume" && a \!== existing?.sessionId);
exec = await runClaudeOnce(fallbackArgs, fallbackConfig.model, fallbackConfig.api, baseEnv);
```

The fallback response still gets returned to the caller but doesn't pollute the primary session.

**Option B: Auto-detect and recover from corruption**

After `execClaude` gets a non-zero exit code, check if stdout contains the signature error pattern. If so, automatically backup and reset the session, then retry:

```typescript
const SIGNATURE_ERROR = /Invalid.*signature.*thinking block/i;

if (exitCode \!== 0 && SIGNATURE_ERROR.test(rawStdout)) {
  console.warn("Detected corrupted session (thinking block signature mismatch), resetting...");
  await backupSession();
  // Retry with fresh session
  return execClaude(name, prompt);
}
```

**Option C: Both A + B**

Implement Option A to prevent future corruption, and Option B as a safety net for sessions that are already corrupted.

**Bonus: Fix the "Unknown" error display**

```typescript
// In forwardToTelegram and forwardToDiscord:
: `error (exit ${result.exitCode}): ${result.stderr || result.stdout || "Unknown"}`
```

### Environment

- Claude Code version: 2.x
- claudeclaw: latest (commit 542410d)
- Scenario: primary model + fallback via different `ANTHROPIC_BASE_URL`

Aleschka:
## Summary

Implements Option C from #18 — both prevention and recovery:

- **Prevention (Option A):** When falling back to an alternative API provider on rate limit, `--resume` and the session ID are stripped from the args. The fallback runs as a one-shot call, keeping the primary session history clean and avoiding thinking block signature mixing.

- **Recovery (Option B):** After any failed execution on a resumed session, stdout+stderr are checked against the signature error pattern. If detected, the corrupted session is automatically backed up and the call is retried with a fresh session.

- **Bonus:** `forwardToTelegram` and `forwardToDiscord` now fall back to `result.stdout` when `result.stderr` is empty, instead of showing "Unknown".

Fixes #18

## Test plan

- [ ] Configure primary + fallback with different `ANTHROPIC_BASE_URL`
- [ ] Trigger rate limit on primary — verify fallback runs without `--resume`
- [ ] Verify primary session remains usable after fallback fires
- [ ] Simulate corrupted session — verify auto-recovery kicks in, backs up old session, creates fresh one
- [ ] Verify Telegram/Discord error messages now show actual error text instead of "Unknown"

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -4,7 +4,7 @@ import { fileURLToPath } from "url";
 import { run, runUserMessage, bootstrap, ensureProjectClaudeMd, loadHeartbeatPromptTemplate } from "../runner";
 import { writeState, type StateData } from "../statusline";
 import { cronMatches, nextCronMatch } from "../cron";
-import { clearJobSchedule, loadJobs } from "../jobs";
+import { clearJobSchedule, loadJobs, snapshotJobFrontmatter } from "../jobs";
 import { writePidFile, cleanupPidFile, checkExistingDaemon } from "../pid";
 import { initConfig, loadSettings, reloadSettings, resolvePrompt, type HeartbeatConfig, type Settings } from "../config";
 import { getDayAndMinuteAtOffset } from "../timezone";
@@ -693,23 +693,30 @@ export async function start(args: string[] = []) {
     const now = new Date();
     for (const job of currentJobs) {
       if (cronMatches(job.schedule, now, currentSettings.timezoneOffsetMinutes)) {
-        resolvePrompt(job.prompt)
-          .then((prompt) => run(job.name, prompt))
-          .then((r) => {
-            if (job.notify === false) return;
-            if (job.notify === "error" && r.exitCode === 0) return;
-            forwardToTelegram(job.name, r);
-            forwardToDiscord(job.name, r);
-          })
-          .finally(async () => {
-            if (job.recurring) return;
-            try {
-              await clearJobSchedule(job.name);
-              console.log(`[${ts()}] Cleared schedule for one-time job: ${job.name}`);
-            } catch (err) {
-              console.error(`[${ts()}] Failed to clear schedule for ${job.name}:`, err);
-            }
-          });
+        snapshotJobFrontmatter(job.name)
+          .then((restoreFrontmatter) =>
+            resolvePrompt(job.prompt)
+              .then((prompt) => run(job.name, prompt))
+              .then(async (r) => {
+                const restored = await restoreFrontmatter();
+                if (restored) {
+                  console.log(`[${ts()}] Restored frontmatter for job: ${job.name}`);
+                }
+                if (job.notify === false) return;
+                if (job.notify === "error" && r.exitCode === 0) return;
+                forwardToTelegram(job.name, r);
+                forwardToDiscord(job.name, r);
+              })
+              .finally(async () => {
+                if (job.recurring) return;
+                try {
+                  await clearJobSchedule(job.name);
+                  console.log(`[${ts()}] Cleared schedule for one-time job: ${job.name}`);
+                } catch (err) {
+                  console.error(`[${ts()}] Failed to clear schedule for ${job.name}:`, err);
+                }
+              })
+          );
       }
     }
     updateState();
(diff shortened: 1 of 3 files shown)


TerrysPOV (changes requested):
This is directionally right for issue #18, but I found two blocking issues that should be addressed before merge:

  1. The fallback path breaks bootstrap / first-run session creation. New sessions start with JSON output so the code can capture session_id, but the fallback path forces text output. The success path still assumes a successful new-session run returns JSON and tries to parse it. If the primary model rate-limits during bootstrap or the first message, fallback can succeed but no session gets created, even though startup still reports the session as live.
  2. This PR also includes the job frontmatter protection changes from PR #25, so it inherits blocking regressions from that branch. As currently written, the frontmatter restore logic can overwrite legitimate concurrent edits and can restore the entire original file when frontmatter parsing fails, which loses body changes.

  To get this mergeable, I think the branch needs to be narrowed and the fallback/session path tightened up. In practice, that likely means:

  - keep the issue #18 fix focused on session/fallback behavior, or rebase/restack it only after the bundled job-frontmatter changes are corrected
  - preserve correct session creation semantics when fallback happens on a brand-new call or during bootstrap
  - ensure resumed-session fallback never pollutes the primary session history
  - keep the corrupted-session recovery path explicit and tested, including backup + fresh-session retry behavior

  Please also add automated coverage for:

  - rate limit on a brand-new session / bootstrap path
  - resumed-session fallback that must not pollute the primary session
  - corrupted-session auto-recovery
  - any bundled src/jobs.ts behavior if those changes remain part of this PR

I was able to run bunx tsc --noEmit successfully in the PR worktree for this branch. I did not find automated tests covering these paths. GitHub also currently marks this PR as CONFLICTING, so it will need to be rebased or otherwise resolved before final validation.

TerrysPOV:
Ported to #159 (fix/job-frontmatter-session-recovery), combined with #25 since both address job reliability. Closing in favour of that PR.

PR #32, opened 2026-03-12
feat: handle text document attachments from Telegram

dmitryanchikov:
## Summary
- Download and serve text documents (code, configs, logs, etc.) to Claude via file path instead of injecting content into the prompt
- Follows the same on-demand pattern used for images — safer, no size limit
- Detects text documents by MIME type and common file extensions
- Also fixes HEARTBEAT_OK check to strip markdown formatting before comparison

## Details
When a user sends a text file (`.py`, `.json`, `.yaml`, `.log`, etc.) via Telegram, the bot now:
1. Downloads the file via Telegram's getFile API to `inbox/telegram/`
2. Passes the file path, name, and size in the prompt
3. Claude reads the file on demand via the Read tool

This avoids injecting potentially large file contents directly into the prompt, which is both safer (no prompt injection from file contents) and more flexible (no size limit).

## Test plan
- [x] Tested with `.txt` file attachment — file downloaded and read correctly
- [x] Tested with `.cookies.txt` — recognized as text, processed correctly
- [ ] Verify non-text documents (images, audio) still handled by existing paths
- [ ] Verify empty/missing documents handled gracefully

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -564,7 +564,8 @@ export async function start(args: string[] = []) {
         })
         .then((r) => {
           if (!r) return;
-          const shouldForward = currentSettings.heartbeat.forwardToTelegram || !r.stdout.trim().startsWith("HEARTBEAT_OK");
+          const stripped = r.stdout.trim().replace(/^[`*_ \n]+/, "");
+          const shouldForward = currentSettings.heartbeat.forwardToTelegram || !stripped.startsWith("HEARTBEAT_OK");
           if (shouldForward) {
             forwardToTelegram("", r);
             forwardToDiscord("", r);
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -208,6 +208,19 @@ function isAudioDocument(document?: TelegramDocument): boolean {
   return Boolean(document?.mime_type?.startsWith("audio/"));
 }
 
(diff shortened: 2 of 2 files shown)


TerrysPOV (changes requested):
This is useful directionally, but I found one blocking regression and one follow-up gap:

  1. As written, this branch narrows Telegram document handling to text-only detection. Current master already supports a broader set of document attachments (PDF / Office docs / markdown / csv) via the existing document-download path, so when you rebase this PR you need to preserve that behavior and extend it, not replace it with isTextDocument() only. Otherwise the rebase will regress attachments that already work on master.
  2. The new extension-based text detection still misses common text/config filenames like Dockerfile, .gitignore, and .env.local when Telegram doesn’t provide a strong MIME type. The .dockerfile entry won’t match Dockerfile with extname().

  I’d suggest:

  - rebase this branch onto current master and preserve the existing document path/commands during conflict resolution
  - fold the text-document improvement into the current broader attachment handling rather than creating a narrower replacement path
  - add tests for text attachments, existing non-text document attachments, and filename-only classification edge cases

I was able to run bunx tsc --noEmit successfully in the PR worktree, but I didn’t find automated coverage for this Telegram attachment flow.

TerrysPOV:
Closing — this feature landed in master via a later PR. Master now has `downloadDocumentFromMessage()` using `isDocumentAttachment()` which handles text documents, and the prompt-injection logic (`documentInfo`, `hasDocument`) is wired into `handleMessage()`. The file-save path, error messaging, and CLAUDE-tool-read instruction are all present.

Thanks @dmitryanchikov for the original implementation — the approach in master follows the same pattern you proposed.

PR #33, opened 2026-03-12
fix: graceful handling of Claude usage limits

dmitryanchikov:
## Summary
- Detect Claude usage limits from CC output and parse the reset time
- Pause heartbeats, cron jobs, and telegram message processing during rate limit
- Send a single notification about the limit (not repeated every interval)
- Reply to Telegram messages with reset time instead of calling Claude
- Auto-resume when the limit resets

## Problem
When Claude hits a usage limit, the daemon kept firing heartbeats and bootstraps every interval, each generating a failed notification. This created notification spam and accumulated failed processes, contributing to OOM on resource-constrained servers.

## Changes
- `runner.ts`: Global rate limit state tracking, reset time parser, exported check functions
- `start.ts`: Rate limit check before heartbeat tick and cron job execution
- `telegram.ts`: Early return with human-readable message when rate-limited

## Test plan
- [ ] Verify rate limit detection from "You've hit your limit · resets Xpm (UTC)" message
- [ ] Verify heartbeats are skipped during rate limit window
- [ ] Verify single Telegram notification sent on limit detection
- [ ] Verify Telegram replies with reset time during limit
- [ ] Verify auto-resume after reset time passes
- [ ] Verify cron jobs are skipped during limit

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -1,7 +1,7 @@
 import { writeFile, unlink, mkdir } from "fs/promises";
 import { join } from "path";
 import { fileURLToPath } from "url";
-import { run, runUserMessage, bootstrap, ensureProjectClaudeMd, loadHeartbeatPromptTemplate } from "../runner";
+import { run, runUserMessage, bootstrap, ensureProjectClaudeMd, loadHeartbeatPromptTemplate, isRateLimited, getRateLimitResetAt, wasRateLimitNotified, markRateLimitNotified } from "../runner";
 import { writeState, type StateData } from "../statusline";
 import { cronMatches, nextCronMatch } from "../cron";
 import { clearJobSchedule, loadJobs } from "../jobs";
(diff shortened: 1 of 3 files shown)


TerrysPOV (changes requested):
This is addressing a real issue, but I found two blocking problems before merge:

  1. The new rate-limit detector is narrower than current master and appears to miss the normal ASCII-apostrophe form of the message ("You've hit your limit"). That means the new pause logic may never activate for the most common output variant.
  2. This branch is based on an older runner.ts / Telegram stack. Since the PR is already CONFLICTING, the rebase/merge resolution needs to preserve the current master thread/session behavior and newer Telegram flow rather than replacing them with the older branch plus the new rate-limit state.

  I’d suggest:

  - rebase onto current master
  - keep the broader existing rate-limit detection and extend it with reset-time parsing rather than narrowing the matcher
  - make sure the final merge preserves current thread-aware runner/session behavior
  - add automated coverage for:
      - ASCII and curly-apostrophe limit messages
      - limit messages without a parseable reset time
      - heartbeat/job suppression during the limit window
      - Telegram short-circuit responses during the limit window
      - automatic resume after reset

I was able to run bunx tsc --noEmit successfully in the PR worktree after installing local deps, but I didn’t find automated tests for these paths. Please consider adding in these automated tests.

TerrysPOV:
Thanks for this contribution, @dmitryanchikov! The rate limit pause logic is exactly the right fix — pausing heartbeats, jobs, and replying to Telegram with a human-readable reset time is a much better experience than notification spam on constrained servers.

This has been ported to a fresh branch against current master as PR #165. Closing this one in favour of the clean port.

PR #34, opened 2026-03-12
feat: send voice messages via Telegram sendVoice API

dmitryanchikov:
## Summary
- Parse `[voice:/path/to/file.ogg]` directives from Claude responses
- Deliver voice files as native Telegram voice messages via `sendVoice` API
- Voice messages sent before any remaining text; if only voice — no empty text sent

## Test plan
- [x] TTS generates `[voice:...]` directive, parsed correctly
- [x] Voice file delivered as Telegram voice message (not plain text)
- [x] Text after voice directive still sent as regular message
- [x] Missing/invalid voice paths silently skipped

### DIFF
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -4,6 +4,7 @@ import { resetSession } from "../sessions";
 import { transcribeAudioToText } from "../whisper";
 import { resolveSkillPrompt, listSkills } from "../skills";
 import { mkdir } from "node:fs/promises";
+import { existsSync } from "node:fs";
 import { extname, join } from "node:path";
 
 // --- Markdown → Telegram HTML conversion (ported from nanobot) ---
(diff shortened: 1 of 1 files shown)


TerrysPOV (changes requested):
This looks reasonable in isolation, but there’s one blocking integration issue before merge:

  1. This branch is behind the current Telegram response pipeline on master. As written, it adds [voice:...] support on top of an older handler that does not include the newer [send-file:...] directive flow. Since GitHub already marks this PR as CONFLICTING, the rebase/merge resolution needs to preserve both behaviors rather than replacing the current handler with the older branch plus voice support.

  I’d suggest:

  - rebase onto current master
  - compose the response pipeline so reactions, voice directives, file directives, and residual text all work together in a defined order
  - add automated coverage for mixed directive responses, especially:
      - [voice:...] only
      - [voice:...] plus normal text
      - [voice:...] plus [send-file:...]
      - invalid/missing voice paths

I also wasn’t able to verify a clean bunx tsc --noEmit in the PR worktree because the branch does not currently have local Bun type dependencies installed. Please install the local Bun type deps in the PR branch and rerun the standard checks there so the rebased branch can be properly verified.

TerrysPOV:
Thanks for this contribution, @dmitryanchikov! The `[voice:...]` directive is a clean and composable approach — it fits naturally into the existing directive pipeline and opens the door for Claude to deliver TTS audio natively to Telegram users.

This has been ported to a fresh branch against current master as PR #166. Closing this one in favour of the clean port.

Issue #14, opened 2026-03-05
[Mega-Post] What should be built next?


moazbuilds:
Hey, thanks for using ClaudeClaw!

I want to know what features matter to you so we can ship things that are actually useful rather than just shipping for the sake of it.

Drop your ideas in the comments, upvote what resonates, and let's go from there..

PeggyFree:
Matrix/Synapse messaging integration

dathtd119:
Im currently using Claudeclaw  with mostly using as remote coding agent. 

So I want to, maybe sync those /slash command to telegram menu, and have some custom command, and able to change mode(Plan, accept Edit, unrestricted, etc.) and granular control when the agent ask for permisson to do something (literally remote control but with Telegram), but just need to only able to use this if enable **CODE_MODE**.

m13v:
A few ideas from building a similar macOS automation agent:

1. **Voice input** - hold a key to dictate instead of typing. Especially useful for describing what you want the agent to do on screen. Local whisper.cpp transcription keeps it fast and private.

2. **Screen context awareness** - capture what is currently on screen (via ScreenCaptureKit or the accessibility tree) so the agent understands your current context without you having to explain it. "Fix the error I'm looking at" instead of copy-pasting the error.

3. **Multi-step macro recording** - let users demonstrate a workflow by clicking through it manually, then have the agent replay and adapt it. The accessibility API gives you enough info to record semantic actions (click button X, type in field Y) rather than pixel coordinates.

4. **Persistent memory** - the agent should remember your preferences across sessions ("always use dark theme", "I prefer pytest over unittest"). An embedding-based memory system makes this searchable without needing exact keyword matches.

5. **Browser automation** - integrating with Playwright for web tasks would cover a huge range of use cases (filling forms, scraping data, navigating dashboards).

robin-ai-builds:
Great tool ! 

I would love to have a local LLM as a backup option, instead of the one using API Keys. 
Reason being even with a small model I can get things done locally, w/o paying extra.

kristian-94:
This tool is really great. It seems like Anthropic has noticed is just going to build their own version? Starting with claude code dispatch, and now channels (discord/telegram): https://code.claude.com/docs/en/channels 

It's not as fully featured as claudeclaw/openclaw yet but it feels like they're going down that path

christian-drescher:
[...]

PR #35, opened 2026-03-12
feat: delegate voice transcription to Claude (remove whisper)

dmitryanchikov:
## Summary
- Remove built-in whisper dependency for voice transcription
- Pass voice file path to Claude with generic instruction to transcribe
- Claude discovers and uses whatever STT tool is available (MCP or otherwise)
- No hardcoded tool names — fully decoupled from specific STT backend

Depends on: #34 (voice sending)

## Test plan
- [x] Voice message received, file downloaded and path passed to Claude
- [x] Claude transcribes using available STT MCP and responds to content
- [x] Works with different STT backends (no hardcoded tool names)
- [x] Graceful fallback message when voice download fails

### DIFF
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -1,7 +1,6 @@
 import { ensureProjectClaudeMd, run, runUserMessage } from "../runner";
 import { getSettings, loadSettings } from "../config";
 import { resetSession } from "../sessions";
-import { transcribeAudioToText } from "../whisper";
 import { resolveSkillPrompt, listSkills } from "../skills";
 import { mkdir } from "node:fs/promises";
 import { extname, join } from "node:path";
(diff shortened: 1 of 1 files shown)


TerrysPOV:
@dmitryanchikov question from my side: what product improvement are you aiming for by removing the built-in Whisper path rather than keeping it as the default fallback? Specifically, what does the pluggable STT approach give users that the current built-in transcription flow does not? Right now it looks like this changes voice from “works out of the box” to “works if the user happens to have some STT-capable tool configured,” so I want to understand the intended upside before I review it as an improvement.

christian-drescher:
Giving you a third perspective from someone who merged this PR into his fork: It makes the product better by keeping the project [lightweight and minimal as per the owners comments](https://github.com/moazbuilds/claudeclaw/issues/14#issuecomment-4263415592).

Anything that can be added as a skill should be left out of the core project. Otherwise, the project gets bloated quickly.

Since the main channel to interact with the bot is Telegram, and Telegram has the voicemail feature, there is an argument for including STT. But, why stopping short of integrating TTS? Hard to draw the line. I suppose whisper is not working on all platforms that are supported by Claude Code. That, and the integrated variant of whisper performing poorly in multi-lingual settings, or with low-resource languages, the user base who actually benefit from STT is probably very small.

Anyhow, there are much better alternatives to whisper these days.

TerrysPOV:
Ported to #158 (feat/stt-delegation) with clean adaptation onto current master. Closing in favour of that PR.

PR #38, opened 2026-03-15
fix: add per-context subprocess timeout to prevent daemon freeze

dmitryanchikov:
## Problem

When a `claude` subprocess hangs indefinitely (e.g. stuck network call, `claude billing` waiting forever), the daemon's serial queue blocks all subsequent messages, heartbeats, and jobs. There is no timeout anywhere in the call chain.

Observed incidents:
- 2026-03-10: simple Telegram message hung 35+ min (exit code 1, empty output)
- 2026-03-14: `claude billing` inside a bash chain hung 33+ min

## Solution

Add a configurable per-context timeout to `runClaudeOnce()` with a SIGTERM → SIGKILL escalation sequence.

**`src/runner.ts`**
- `runClaudeOnce()` accepts a `timeoutMs` parameter
- On expiry: sends SIGTERM, then SIGKILL after a 5s grace period
- The kill causes the process's pipes to close, so `Response.text()` and `proc.exited` resolve naturally
- Fallback model retry is skipped on timeout (only meaningful for rate limits)
- Log files record `[TIMED OUT]` for observability

**`src/config.ts`**
- New `TimeoutsConfig` interface added to `Settings`
- Timeouts are read via `getSettings()` on every invocation — changes to `settings.json` take effect within the daemon's existing 30s hot-reload cycle, **no restart required**

## Default timeouts

| Context | Default |
|---------|----------|
| `telegram` | 5 min |
| `heartbeat` | 15 min |
| jobs / other | 30 min |

All configurable in `settings.json`:
```json
"timeouts": {
  "telegram": 5,
  "heartbeat": 15,
  "job": 30,
  "default": 5
}
```

### DIFF
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -623,7 +623,11 @@ async function handleMessage(message: TelegramMessage): Promise<void> {
     const result = await runUserMessage("telegram", prefixedPrompt);
 
     if (result.exitCode !== 0) {
-      await sendMessage(config.token, chatId, `Error (exit ${result.exitCode}): ${result.stderr || "Unknown error"}`, threadId);
+      const isKilled = result.exitCode === 143 || result.exitCode === 137;
+      const errorMsg = isKilled
+        ? `⏱ Request timed out (exit ${result.exitCode}: ${result.exitCode === 143 ? "SIGTERM" : "SIGKILL"}) — the subprocess took too long and was killed. Try again or split into smaller steps.`
+        : `Error (exit ${result.exitCode}): ${result.stderr || "Unknown error"}`;
+      await sendMessage(config.token, chatId, errorMsg, threadId);
     } else {
       const { cleanedText, reactionEmoji } = extractReactionDirective(result.stdout || "");
       if (reactionEmoji) {
--- a/src/config.ts
+++ b/src/config.ts
@@ -25,10 +25,11 @@ const DEFAULT_SETTINGS: Settings = {
     forwardToTelegram: true,
   },
   telegram: { token: "", allowedUserIds: [] },
-  discord: { token: "", allowedUserIds: [], listenChannels: [] },
+  discord: { token: "", allowedUserIds: [] },
   security: { level: "moderate", allowedTools: [], disallowedTools: [] },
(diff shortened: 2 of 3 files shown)


TerrysPOV (changes requested):
This is a good problem to solve, but I found two blocking issues before merge:

  1. This branch is based on an older config/runner stack and drops unrelated current-master behavior, including discord.listenChannels config support and newer runner/session features. Since the PR is already CONFLICTING, the rebase/merge resolution needs to preserve the current master config/runtime behavior rather than replacing it with the older branch plus timeout handling.
  2. The new timeout config exposes timeouts.default, but the implementation never uses it. Right now every non-telegram/non-heartbeat invocation falls through to the job timeout, so the documented default setting is effectively dead.

  I’d suggest:

  - rebase onto current master
  - preserve current master config fields and thread/session runner behavior during conflict resolution
  - either wire timeouts.default into the fallback path for non-telegram/non-heartbeat/non-job invocations, or remove it from the config surface and docs
  - add automated coverage for:
      - hung telegram subprocess timing out and releasing the queue
      - heartbeat/job timeout behaviour
      - timeout without fallback retry
      - rebased config parsing preserving existing fields like discord.listenChannels

I also wasn’t able to verify a clean bunx tsc --noEmit in the PR worktree because local Bun type deps are not installed there, so please install those in the PR branch and rerun the standard checks after rebasing.

TerrysPOV:
Closing in favour of #155 (feat/per-context-timeouts), a clean port targeting current master.

**What was preserved vs the original:**
- `TimeoutsConfig` interface and per-context timeout values (telegram/heartbeat/job/default defaults: 5/15/30/5 min)
- `resolveTimeoutMs(name)` mapping invocation name to the correct timeout
- Rate limit state tracking: `parseRateLimitResetTime()`, `isRateLimited()`, `getRateLimitResetAt()`
- Timeout exit code updated to `124` (master's convention; original used `143`/`137`)

**What was NOT ported (broke master):**
- The original removed `discord.listenChannels` from config — master depends on it; preserved in #155

Thanks @nabilfatih for the original work. All attribution is in #155.

PR #43, opened 2026-03-17
feat: session auto-rotation with optional summary generation

artemasmith:
## Problem

Long-running ClaudeClaw sessions degrade as context fills up — Claude starts contradicting itself and producing lower quality responses. The built-in compression only kicks in near the context limit, not preventing gradual degradation.

## Solution

Automatic session rotation based on configurable thresholds:

- **Message count** — rotates after N messages (default: 50)
- **Session age** — rotates after N hours (default: 24)

When `summaryPath` is configured, the expiring session generates a markdown summary before rotating. The latest summary is loaded into the bootstrap prompt of the new session, preserving context continuity.

## Changes

- `src/sessions.ts` — added `messageCount` to `GlobalSession`, incremented on each `getSession()`
- `src/config.ts` — added `SessionConfig` interface and parsing
- `src/rotation.ts` — new file: `needsRotation()`, `rotateSession()`, `loadLatestSummary()`
- `src/runner.ts` — rotation check before `execClaude()`, summary loading in `bootstrap()`
- `prompts/SUMMARY.md` — prompt template for summary generation

## Configuration

New `session` section in `settings.json`:

```json
{
  "session": {
    "autoRotate": true,
    "maxMessages": 50,
    "maxAgeHours": 24,
    "summaryPath": "/path/to/summaries"
  }
}
```

All fields are optional with sensible defaults. When `summaryPath` is empty (default), rotation happens without summary generation.

## Backward Compatibility

- `messageCount` defaults to `0` for existing sessions via `|| 0` fallback
- Feature is opt-in: `autoRotate` defaults to `true` but does nothing without the config section
- No changes to existing behavior when `session` config is absent


TerrysPOV (changes requested):
This is solving a real problem, but I found several blocking issues before merge:

  1. The branch does not currently typecheck, and part of the failure is in the new feature itself: messageCount is used in the rotation code and in sessions.ts, but GlobalSession is not updated to include it.
  2. maxMessages is wired to the wrong signal. Right now messageCount is incremented inside getSession(), which is used for many things besides actual user/assistant turns. That means rotation would be driven by arbitrary session reads rather than real conversation message count.
  3. This branch is built on an older global-only runner and does not integrate with current thread-session behavior on master. Since the PR is already CONFLICTING, the rebase/merge resolution needs to preserve current thread-aware runner/session behavior, and the rotation design needs to say explicitly what should happen for thread sessions.
  4. The summary path introduces an unexpected language change: prompts/SUMMARY.md and the bootstrap handoff text are hardcoded in Russian. That feels like an unintended product behaviour change for the default repo experience.

  I’d suggest:

  - rebase onto current master
  - make the new counter explicit and type-safe
  - count actual conversation turns/messages, not getSession() calls
  - define whether rotation applies only to the global session or also to thread sessions, and implement that consistently
  - replace the Russian-only summary/bootstrap text with repo-appropriate default language or make it configurable
  - add automated coverage for:
      - rotation threshold reached on real message count
      - age-based rotation
      - summary generation success/failure
      - no accidental rotation from status/compact/session inspection calls
      - thread-session behavior after rebasing

I wasn’t able to verify a clean bunx tsc --noEmit on the branch as submitted because of the type errors above. Please fix the type errors and then consider introducing automated test for bunx tsc.

TerrysPOV:
Thanks @artemasmith — this has been picked up, rebased onto current master, and cleaned up with a few fixes (English summary prompt, cleanSpawnEnv parity, interface declaration, opt-in default). New PR open at #149. Closing this in favour of it.

PR #45, opened 2026-03-17
feat: sync daemon session with active Claude Code terminal windows

### LINKED ISSUE #39: Sync deamon with all chat windows
When I have a claudeClaw instance open on my computer through claude code and I also send some telegram messages, the two are out of sync, and one doesn't know about the work the other is doing. It would be nice if these are synced together to get a more coherent experience switching between telegram - claude code

alexflorisca:
## Summary

Fixes #39

When a user is actively chatting in a Claude Code terminal session while the daemon is also serving Telegram or Discord, the two windows are disconnected — each has its own conversation context and neither knows what the other is doing.

This PR adds automatic terminal session detection:

- Before each execution, claudeclaw scans `~/.claude/projects/<project>/` for `.jsonl` session files (Claude Code's native session storage)
- If the most recently modified session file is newer than claudeclaw's own `lastUsedAt` timestamp, claudeclaw adopts that session ID
- All subsequent Telegram/Discord messages then `--resume` the same thread the user was just in from their terminal, and vice versa
- On daemon startup, if a terminal session already exists, bootstrap skips creating a new session and joins the active one instead

The sync is transparent — a log line is printed when a session switch occurs, and no configuration is needed.

## Test plan

- [ ] Start a Claude Code terminal session in a project (`claude` or `claude --resume`)
- [ ] Start claudeclaw daemon in the same project directory
- [ ] Send a Telegram/Discord message — confirm the daemon continues the terminal conversation context
- [ ] Do work in the terminal after the daemon runs — confirm the next Telegram message picks it back up
- [ ] Verify daemon log shows `Session sync: adopting terminal session <id>...` when a switch occurs
- [ ] Verify no change in behaviour when no terminal session exists (fresh project)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 node_modules
 .codemachine
 AGENTS.md
+.worktrees
--- a/src/sessions.ts
+++ b/src/sessions.ts
@@ -1,5 +1,6 @@
 import { join } from "path";
-import { unlink, readdir, rename } from "fs/promises";
+import { unlink, readdir, rename, stat } from "fs/promises";
+import { homedir } from "os";
 
 const HEARTBEAT_DIR = join(process.cwd(), ".claude", "claudeclaw");
 const SESSION_FILE = join(HEARTBEAT_DIR, "session.json");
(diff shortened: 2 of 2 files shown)


TerrysPOV:
This does not match the project’s intended session model. I confirmed with the repo owner that channel separation is intrinsic to ClaudeClaw, so terminal/TUI should remain isolated by default rather than being implicitly merged into the
daemon’s global session. I am therefore closing the PR.

PR #50, opened 2026-03-20
Adds more flexibility for heartbeats to place HEARTBEAT_OK

christian-drescher:
Heartbeat messages are only suppressed in the current master when the agent starts its response with HEARTBEAT_OK, potentially followed with other text (e.g., justifications or internal thoughts).

Quite often, my agent tends to **end** heartbeats with HEARTBEAT_OK rather than starting them. Hence, the heartbeat message is forwarded to me rather than suppressed (albeit the agent intended to suppress).

This PR generalizes the current handling to suppress any heartbeat message that contains the string HEARTBEAT_OK.

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -564,7 +564,7 @@ export async function start(args: string[] = []) {
         })
         .then((r) => {
           if (!r) return;
-          const shouldForward = currentSettings.heartbeat.forwardToTelegram || !r.stdout.trim().startsWith("HEARTBEAT_OK");
+          const shouldForward = currentSettings.heartbeat.forwardToTelegram || !r.stdout.includes("HEARTBEAT_OK");
           if (shouldForward) {
             forwardToTelegram("", r);
             forwardToDiscord("", r);
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -564,7 +564,9 @@ export async function start(args: string[] = []) {
         })
         .then((r) => {
           if (!r) return;
-          const shouldForward = currentSettings.heartbeat.forwardToTelegram || !r.stdout.includes("HEARTBEAT_OK");
+          const normalized = r.stdout.trim();
+          const shouldSuppress = normalized.startsWith("HEARTBEAT_OK") || normalized.endsWith("HEARTBEAT_OK");
+          const shouldForward = currentSettings.heartbeat.forwardToTelegram || !shouldSuppress;
           if (shouldForward) {
             forwardToTelegram("", r);
             forwardToDiscord("", r);


pbrisson:
I'm noticing similar behaviour as well. The heartbeats often slips in my Discord/Telegram for this specific reason.

TerrysPOV (changes requested):
I see the problem this is trying to solve, but I think the new match is a bit too broad.

Right now the prompt defines HEARTBEAT_OK as the sentinel no-op reply. Changing the suppression logic from startsWith("HEARTBEAT_OK") to includes("HEARTBEAT_OK") means any heartbeat output that merely mentions that token anywhere will now be suppressed, even if it also contains real user-facing content.

So this fixes the “agent ends with HEARTBEAT_OK” case, but it also creates a new false-positive suppression case.

I’d suggest narrowing the match instead of using a raw substring check, for example:

  - suppress when the normalized output is exactly HEARTBEAT_OK
  - or suppress when the output contains HEARTBEAT_OK plus only surrounding whitespace / trivial wrapper text
  - but avoid suppressing arbitrary messages just because they mention the token somewhere

I was able to run bun install and bunx tsc --noEmit successfully in the PR worktree.

christian-drescher:
>- suppress when the normalized output is exactly HEARTBEAT_OK

Well, this is actually a tighter requirement than *starting* the response with HEARTBEAT_OK. It would make the experience worse.

>- or suppress when the output contains HEARTBEAT_OK plus only surrounding whitespace / trivial wrapper text

Clearing surrounding whitespace is what the project already does. Trivial wrapper text is best classified by using an LLM-based agent - we don't want to go down this route.

>- but avoid suppressing arbitrary messages just because they mention the token somewhere

Because we want to allow the agent to say HEARTBEAT_OK in cases where HEARTBEAT is actually not okay? (Friendly sarcasm.)

What if we suppressed Telegram forwarding on HEARTBEAT_OK at *start* **or** *end* of any heartbeat response (after trimming whitespace)?

TerrysPOV:
I think trim() + suppress when HEARTBEAT_OK is at the start or end of the response is the right middle ground here. It fixes the end-of-response case this PR is targeting without broadening suppression to any message that merely mentions the token somewhere in the middle. That keeps the sentinel behavior reasonably strict while avoiding the false positives introduced by includes().

  If you want a concrete condition, something like:

```
	const normalized = output.trim();
  	const shouldSuppress =
    	normalized.startsWith("HEARTBEAT_OK") ||
    	normalized.endsWith("HEARTBEAT_OK");
```

That is much easier to justify in review than unqualified substring matching.

TerrysPOV (approved):
Nice work!

I don’t see a blocking issue in the updated version. The new trim() + startsWith("HEARTBEAT_OK") || endsWith("HEARTBEAT_OK") rule looks like the right compromise here: it fixes the end-of-response case this PR is targeting without broadening suppression to any message that merely mentions the token somewhere in the middle.

I was able to run bun install and bunx tsc --noEmit successfully on the updated branch.

PR #55, opened 2026-03-23
fix: add missing discordUserIds param and settings variable

alanfeng99:
## Summary
- **config.ts**: `parseSettings` was called with a `discordUserIds` argument from `loadSettings`/`reloadSettings`, but the function signature only accepted one parameter (`raw`). This caused a `ReferenceError: discordUserIds is not defined` at startup.
- **runner.ts**: `execClaude` destructured `getSettings()` directly into named variables but later referenced the full `settings` object for `sessionTimeoutMs`, causing a `ReferenceError: settings is not defined` during session bootstrap.

Both bugs are crash-on-start — the daemon cannot launch without these fixes.

## Test plan
- [ ] Run `bun run src/index.ts start --web` and verify daemon starts without ReferenceError
- [ ] Verify Discord user ID parsing still works with snowflake IDs (large numbers)
- [ ] Verify `sessionTimeoutMs` setting is respected when configured

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,7 +1,10 @@
 import { join, isAbsolute } from "path";
 import { mkdir } from "fs/promises";
 import { existsSync } from "fs";
-import { normalizeTimezoneName, resolveTimezoneOffsetMinutes } from "./timezone";
+import {
+  normalizeTimezoneName,
+  resolveTimezoneOffsetMinutes,
+} from "./timezone";
 
 const HEARTBEAT_DIR = join(process.cwd(), ".claude", "claudeclaw");
 const SETTINGS_FILE = join(HEARTBEAT_DIR, "settings.json");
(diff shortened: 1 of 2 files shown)


TerrysPOV (changes requested):
I do think there is a real bug here, but this branch needs to be rebased/narrowed before it’s mergeable.

  1. The parseSettings / discordUserIds fix looks valid and addresses a real current-master issue: loadSettings() and reloadSettings() pass a second argument, but parseSettings() does not currently declare it.
  2. However, this PR is built on an older runner/session stack. The branch’s runner.ts is still global-session-only and does not include the current master thread-aware/session-manager behavior. Since the PR is already CONFLICTING, the rebase/merge resolution needs to preserve current master runner/session behavior rather than replacing it with this older branch.
  3. The PR description also mentions a settings is not defined runner crash, but that part appears to already be fixed on current master. So the actionable current-master fix here is narrower than the PR description suggests.

  I’d suggest:

  - rebase onto current master
  - keep the parseSettings(discordUserIds) fix
  - drop any stale runner changes that are already fixed upstream or that would regress current thread/session behavior
  - update the PR description/test plan so it reflects the actual remaining fix on top of current master

I was able to run bunx tsc --noEmit successfully in the PR worktree after installing local deps, but the PR is currently CONFLICTING and git merge-tree shows conflicts in src/config.ts, so final validation needs to happen on a rebased branch.

TerrysPOV:
Both fixes in this PR are already present in master:

1. The `discordUserIds` parameter added to `parseSettings()` (line 251 of `src/config.ts`) — already merged.
2. `sessionTimeoutMs` in the Settings interface and DEFAULT_SETTINGS — already present.

Closing in favour of the already-merged changes. Thanks @alanfeng99 for identifying these issues.

PR #58, opened 2026-03-24
feat: add shared messaging utils and surface real Claude CLI errors

sean808080:
## Problem

Claude CLI exits non-zero on auth and quota failures, but the human-readable error message (e.g. `Not logged in · Please run /login`) lives in the JSON stdout `result` field — **not** in stderr. All three chat bridges were showing `Unknown error` instead of the actual failure reason.

**Before:**
```
Error (exit 1): Unknown error
```
**After:**
```
Error (exit 1): Not logged in · Please run /login
```

## Solution

Introduces `src/messaging.ts` — a shared utilities module used by `start.ts`, `telegram.ts`, and `discord.ts`.

### Key function: `extractErrorDetail(result)`

```typescript
export function extractErrorDetail(result: { stdout: string; stderr: string }): string {
  const stderr = result.stderr?.trim() || "";
  if (stderr) return stderr;

  // Claude CLI emits human-readable errors in JSON stdout when stderr is empty
  try {
    const parsed = JSON.parse(stdout);
    if (parsed?.is_error && parsed?.result) return String(parsed.result).trim();
    if (parsed?.error?.message) return String(parsed.error.message).trim();
    if (typeof parsed?.error === "string") return parsed.error.trim();
  } catch {}

  return stdout;
}
```

### Additional utilities (no behaviour change, available for future use)

- `describeProvider()` — human-readable provider/model label
- `isProviderStatusQuery()` / `isModelChangeRequest()` — intent detection for model-switch queries
- `authoritativeProviderReply()` — canonical provider status response
- `extractReactionDirective()` — strips `[react:emoji]` directives from response text

### Files changed

| File | Change |
|------|--------|
| `src/messaging.ts` | New shared utilities module |
| `src/commands/start.ts` | Use `extractErrorDetail` in `forwardToTelegram` and `forwardToDiscord` |
| `src/commands/telegram.ts` | Use `extractErrorDetail` in the main message handler error path |
| `src/commands/discord.ts` | Use `extractErrorDetail` in the main message handler error path |

## Testing

Auth failures now surface the real Claude CLI message across all three chat interfaces. The fix handles:
- `stderr` present → returns stderr (existing behaviour, unchanged)
- `stderr` empty, JSON stdout with `is_error: true` → returns `result` field  
- `stderr` empty, JSON stdout with `error.message` → returns `error.message`
- Fallback to raw stdout if JSON parse fails

### DIFF
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -1,5 +1,6 @@
 import { ensureProjectClaudeMd, run, runUserMessage, compactCurrentSession } from "../runner";
 import { getSettings, loadSettings } from "../config";
+import { extractErrorDetail } from "../messaging";
 import { resetSession, peekSession } from "../sessions";
 import { readFile } from "node:fs/promises";
 import { existsSync } from "node:fs";
@@ -479,7 +480,7 @@ async function handleMessageCreate(token: string, message: DiscordMessage): Prom
     const result = await runUserMessage("discord", prefixedPrompt);
 
     if (result.exitCode !== 0) {
-      await sendMessage(config.token, channelId, `Error (exit ${result.exitCode}): ${result.stderr || "Unknown error"}`);
+      await sendMessage(config.token, channelId, `Error (exit ${result.exitCode}): ${extractErrorDetail(result) || "Unknown error"}`);
     } else {
       const { cleanedText, reactionEmoji } = extractReactionDirective(result.stdout || "");
       if (reactionEmoji) {
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -2,6 +2,7 @@ import { writeFile, unlink, mkdir } from "fs/promises";
 import { join } from "path";
 import { fileURLToPath } from "url";
 import { run, runUserMessage, bootstrap, ensureProjectClaudeMd, loadHeartbeatPromptTemplate } from "../runner";
+import { extractErrorDetail } from "../messaging";
 import { writeState, type StateData } from "../statusline";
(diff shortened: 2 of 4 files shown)


TerrysPOV (changes requested):
I think the core idea here is good, but there are two issues before I’d treat this as mergeable:

  1. The PR branch does not currently typecheck, and the failures are unrelated existing config.ts / runner.ts startup issues (discordUserIds / parseSettings / settings undefined). Since the PR is also CONFLICTING, I can’t properly validate this fix on the branch as submitted. Please rebase onto current master (or otherwise include the startup-fix branch) first so this can be verified on a working branch.
  2. src/messaging.ts is much broader than the actual change. In this PR, only extractErrorDetail() is used; the other exported helpers are dead code, and extractReactionDirective() is still duplicated locally in Telegram and Discord instead of actually being shared. I’d strongly prefer narrowing this PR to the error-detail helper you actually need right now.

  I’d suggest:

  - rebase onto current master / a branch that already includes the config.ts and runner.ts startup fixes
  - keep extractErrorDetail() if that’s the real fix
  - drop the unused provider-status / reaction helpers from this PR unless they are being wired in now
  - add coverage for non-zero Claude CLI responses where:
      - stderr is populated
      - stderr is empty but stdout contains JSON with is_error/result
      - stdout is non-JSON fallback text

I wasn’t able to verify a clean bunx tsc --noEmit on the PR branch as submitted because of the existing startup/type errors above.

sean808080:
i'm off claudeclaw and onto Hermes.  Good luck.

PR #62, opened 2026-03-26
fix: prevent double Telegram polling loop on startup

### LINKED ISSUE #47: Telegram 409 Conflict — double polling loop on startup
## Bug

The daemon opens two concurrent TCP connections to `api.telegram.org` on startup, causing `getUpdates` to return `409 Conflict` on every poll. Messages still get through via retries, but logs are spammed indefinitely.

## Evidence

```
$ ss -tnp | grep 149.154
ESTAB  pid=XXXX,fd=17  → api.telegram.org:443
ESTAB  pid=XXXX,fd=18  → api.telegram.org:443
```

Single daemon process, two connections. Only one should exist for polling.

## Reproduction

- 3 agents running identical code, same Bun version, same plugins
- Only one agent consistently has the issue
- Clean restart (stop → delete session.json + daemon.pid → wait 35s → start) does **not** fix it
- Stopping the daemon and calling `getUpdates` manually returns `200 OK` — conflict is internal

## Root cause (suspected)

`startPolling()` in `src/commands/telegram.ts` has no guard against being called twice:

```typescript
export function startPolling(debug = false): void {
  telegramDebug = debug;
  (async () => {
    await ensureProjectClaudeMd();
    await poll();
  })().catch((err) => {
    console.error(`[Telegram] Fatal: ${err}`);
  });
}
```

If `initTelegram()` is invoked twice before the first `poll()` enters its `while` loop, two polling loops run concurrently.

## Suggested fix

```typescript
let isPolling = false;

export function startPolling(debug = false): void {
  if (isPolling) return;
  isPolling = true;
  // ...
}
```

## Environment

- ClaudeClaw v1.0.0
- Bun 1.3.10
- Debian 12

bluzername:
## Problem

The Telegram bot daemon opens two concurrent connections to `api.telegram.org` on startup, which causes `getUpdates` to return `409 Conflict` on every poll. Messages still work through retries but the logs are spammed with error messages nonstop.

I had this happening on one of my agents and the logs were full of 409 errors which made it hard to see the real activity.

## Root Cause

`startPolling()` in `src/commands/telegram.ts` has no guard against being called twice. If `initTelegram()` is invoked twice before the first `poll()` enters its `while (running)` loop, two polling loops start concurrently - each opening its own TCP connection to Telegram API.

## Fix

Added an `isPolling` flag that prevents `startPolling()` from running more than once:

```typescript
let isPolling = false;

export function startPolling(debug = false): void {
  if (isPolling) return;  // already running, skip
  isPolling = true;
  // ...
}
```

The flag is also reset to `false` in the catch handler so if the first attempt fails completely, a retry can still work.

This is the exact fix suggested in the issue. Simple, no side effects, covers the race condition.

Closes #47

### DIFF
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -952,6 +952,7 @@ async function registerBotCommands(token: string): Promise<void> {
 // --- Polling loop ---
 
 let running = true;
+let isPolling = false;
 
 async function poll(): Promise<void> {
   const config = getSettings().telegram;
@@ -1030,12 +1031,15 @@ process.on("SIGINT", () => { running = false; });
 
 /** Start polling in-process (called by start.ts when token is configured) */
 export function startPolling(debug = false): void {
+  if (isPolling) return;
+  isPolling = true;
   telegramDebug = debug;
   (async () => {
     await ensureProjectClaudeMd();
     await poll();
   })().catch((err) => {
     console.error(`[Telegram] Fatal: ${err}`);
+    isPolling = false;
   });
 }
 


TerrysPOV (approved):
I don’t see a blocking issue in this change. The isPolling guard looks like a reasonable fix for preventing duplicate in-process Telegram polling loops on startup, and the flag reset in the fatal catch preserves retry behavior if startup fails before polling is established.

I wasn’t able to verify a clean bunx tsc --noEmit on the branch, but the failures are on separate existing config.ts / runner.ts issues unrelated to this PR. The only follow-up risk I’d note is that isPolling is not reset on graceful shutdown, so this assumes Telegram polling is not expected to stop and restart within the same long-lived process. There is no evidence that this is the case. PR merged.

PR #64, opened 2026-03-26
fix(config): add missing discordUserIds param to parseSettings

### LINKED ISSUE #63: bug: parseSettings missing discordUserIds parameter causes ReferenceError on startup
## Bug Description

`parseSettings()` in `src/config.ts` (line 220) is missing the `discordUserIds` parameter from its signature, but:

1. Both callers (`loadSettings` at line ~340 and `reloadSettings` at line ~348) pass a second argument: `parseSettings(raw, extractDiscordUserIds(rawText))`
2. The function body (line ~252) references `discordUserIds` to populate `discord.allowedUserIds`

This causes a `ReferenceError: discordUserIds is not defined` crash on daemon startup.

```
ReferenceError: discordUserIds is not defined
      at parseSettings (src/config.ts:252:23)
      at loadSettings (src/config.ts:340:12)
```

## Root Cause

An incomplete refactoring in commit `58f6c4f` removed `discordUserIds` from the function signature but left the body and callers unchanged.

## Why This Matters

Discord snowflake IDs exceed `Number.MAX_SAFE_INTEGER` (2^53 - 1). The `extractDiscordUserIds()` function exists specifically to extract these IDs as raw strings from the JSON text before `JSON.parse` loses precision. Without this parameter, Discord authentication silently breaks even if the crash is worked around.

## Fix

One-line change — add the parameter back to the signature:

```diff
-function parseSettings(raw: Record<string, any>): Settings {
+function parseSettings(raw: Record<string, any>, discordUserIds?: string[]): Settings {
```

## Related

- PR #55 addresses this among other changes

bladewing:
## Summary
- Restores the `discordUserIds?: string[]` parameter to `parseSettings()` in `src/config.ts` (accidentally removed in `58f6c4f`), fixing `ReferenceError: discordUserIds is not defined` from `loadSettings`/`reloadSettings`.
- Declares `const settings = getSettings()` in `execClaude` (`src/runner.ts`) so the existing `(settings as any).sessionTimeoutMs` reference resolves. Previously `execClaude` only destructured `getSettings()`, leaving `settings` undefined and causing a `ReferenceError` on every Discord message / heartbeat / job run (and a `tsc --noEmit` failure).

Both bugs are crash-on-execute caused by incomplete refactors that left dangling references; both fixes are minimal and mirror patterns already used elsewhere in the file (`compactCurrentSession` declares `settings` the same way).

## Context
- Discord snowflake IDs exceed `Number.MAX_SAFE_INTEGER`. `extractDiscordUserIds()` extracts them as raw strings before `JSON.parse` loses precision — restoring the `discordUserIds` parameter keeps that pipeline working.
- `compactCurrentSession` already does `const settings = getSettings(); … settings.sessionTimeoutMs …`. We use the same pattern in `execClaude`.
- The broader `sessionTimeoutMs`-through-`parseSettings` cleanup (removing the `as any` cast) is intentionally **not** done here — that's the scope of #101.

Closes #63

## Test plan
- [x] `bunx tsc --noEmit` is no longer blocked on `settings is not defined` in `src/runner.ts` (two remaining errors are pre-existing in `src/ui/server.ts` and `src/whisper.ts`, unrelated to this PR).
- [x] `bun run src/index.ts start --web` starts without ReferenceError.
- [ ] Configure Discord with snowflake user IDs and verify they are parsed correctly as strings.
- [ ] Send a Discord message / trigger a heartbeat to exercise `execClaude` and confirm no `ReferenceError: settings is not defined`.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.0",
+      "version": "1.0.1",
       "keywords": ["cron", "heartbeat", "scheduler", "daemon"],
       "category": "productivity"
     }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/config.ts
+++ b/src/config.ts
@@ -217,7 +217,7 @@ function parseAgenticConfig(raw: any): AgenticConfig {
   };
 }
 
-function parseSettings(raw: Record<string, any>): Settings {
+function parseSettings(raw: Record<string, any>, discordUserIds?: string[]): Settings {
   const rawLevel = raw.security?.level;
   const level: SecurityLevel =
     typeof rawLevel === "string" && VALID_LEVELS.has(rawLevel as SecurityLevel)
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -336,7 +336,8 @@ async function execClaude(name: string, prompt: string): Promise<RunResult> {
   const timestamp = new Date().toISOString().replace(/[:.]/g, "-");
   const logFile = join(LOGS_DIR, `${name}-${timestamp}.log`);
 
-  const { security, model, api, fallback, agentic } = getSettings();
+  const settings = getSettings();
+  const { security, model, api, fallback, agentic } = settings;
 
   // Determine which model to use based on agentic routing
   let primaryConfig: ModelConfig;


TerrysPOV (changes requested):
This one-line fix looks correct to me. Restoring discordUserIds?: string[] on parseSettings() matches the existing function body and the two existing callers, and it restores the intended raw-string handling for Discord snowflake IDs.

One note on validation: I could not confirm the broader “daemon starts cleanly” claim from the test plan because the branch still fails bunx tsc --noEmit on a separate existing src/runner.ts issue (settings is undefined there). So this
PR appears to fix the parseSettings startup crash, but it does not by itself get the branch to a clean startup/typecheck state. It would be good to narrow the PR description/test plan to that specific fix, or stack it with the separate
runner fix if the goal is a fully clean startup path. Once you decidw which approach to take and make the appropriate changes, I'll re-review and merge if it looks good.

bladewing:
@TerrysPOV thanks for the careful review.

Went with option B and stacked the runner fix here so the branch is typecheck-clean. Specifically: `execClaude` was destructuring `getSettings()` directly, so `(settings as any).sessionTimeoutMs` at the bottom of the function was referencing an undeclared `settings`. New commit `e57e9ee` adds the missing `const settings = getSettings();` and destructures from it — same pattern `compactCurrentSession` already uses, so it's a 2-line minimum-diff fix.

I deliberately kept the `as any` cast in place. Removing it (and adding `sessionTimeoutMs` to the `Settings` type) is the scope of #101, and I didn't want to overlap. Re-requesting review.

bladewing:
CI note: pushed `15a97ad` to bump `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` to `1.0.1` — `marketplace-version-guard` and `plugin-version-guard` now pass.

`claude-review` is failing on a runner-side Anthropic quota error (`Claude Code returned an error result: You're out of extra usage · resets 5:30pm (UTC)`), unrelated to this PR's diff. Should clear on the next re-run after the quota resets.

TerrysPOV (approved):
I rechecked the full current PR tip, and I don’t see a blocking issue here.

The branch is properly scoped to the intended statusline rendering fix plus the expected plugin/marketplace version bumps from the new guard workflow. The render()-based approach in src/commands/start.ts looks like the right fix: it strips ANSI before width calculation, measures visible content width dynamically, and pads the content row so the header, body, and footer stay aligned when Discord/Telegram indicators are present.

I wasn’t able to verify a clean bunx tsc --noEmit on the branch as submitted, but the remaining failure is still the separate existing parseSettings signature issue in src/config.ts, not this statusline change. So from my side, the actual change here looks merge-safe.

TerrysPOV:
Superseeded by 119 and merged aspart of 121

PR #67, opened 2026-03-30
fix: prevent memory leaks in runner that cause OOM over time

archon-agent:
## Summary

Three fixes for memory leaks in `src/runner.ts` that cause the bun process to grow unboundedly over time, eventually triggering OOM kills on long-running daemon deployments.

### 1. Capped stream collection (10 MB limit)
- Replaced unbounded `new Response(proc.stdout).text()` with a `collectStream()` helper that caps output at 10 MB
- Prevents runaway or very large Claude responses from consuming all available RAM

### 2. Timer cleanup after process exit
- The `setTimeout` for process timeout was never cleared on normal completion
- Added `clearTimeout` in both success and error paths to prevent timer reference leaks

### 3. Queue chain reference reset
- `task.catch(() => {})` passes through fulfilled values, keeping references to all previous task results in the promise chain
- Changed to `task.then(() => {}, () => {})` which discards result references, allowing GC to reclaim them
- Applied to both `globalQueue` and `threadQueues`

## Context

In production (24/7 daemon serving Telegram bot), these leaks caused the bun process to grow from ~43 MB to **23 GB** over several days, eventually triggering an OOM kill and crashing the VM.

## Test plan
- [x] Verified fixes on a production VM200 instance running claudeclaw as a systemd daemon
- [x] After applying fixes, daemon memory stayed stable at ~43 MB over extended operation
- [ ] No existing tests to run, but changes are minimal and isolated to resource management

### DIFF
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -61,6 +61,8 @@ const RATE_LIMIT_PATTERN = /you.ve hit your limit|out of extra usage/i;
 
 // Serial queue — prevents concurrent --resume on the same session
 // Global queue for non-thread messages (backward compatible)
+// Reset to a fresh resolved promise after each task to avoid holding
+// references to every previous result (memory leak).
 let globalQueue: Promise<unknown> = Promise.resolve();
 // Per-thread queues — each thread runs independently in parallel
 const threadQueues = new Map<string, Promise<unknown>>();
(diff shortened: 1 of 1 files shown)


TerrysPOV (changes requested):
I found one blocking issue in the new stream-capping logic:

  1. collectStream() stops reading stdout/stderr once the 10 MB cap is reached, but runClaudeOnce() clears the timeout before awaiting proc.exited. If the Claude child keeps writing after the cap is hit, it can block on a full pipe and never exit, and there is no timeout left to recover. That means a runaway response can hang the invocation indefinitely rather than just bounding memory use.

The timer cleanup and queue-chain reset both look good to me. I’d suggest keeping those, but changing the output cap so the process is still drained or explicitly terminated once the cap is exceeded, rather than just stopping reads and waiting for normal exit.

I was able to assess the runner changes themselves, but I couldn’t verify a clean bunx tsc --noEmit on the branch because of the separate existing src/config.ts parseSettings signature issue. This PR should be rebased on top of #64 (or equivalent) before final validation.

squirblej:
Addressed in #142. The  now drains rather than breaks after the cap — pipe stays clear, child exits naturally, clearTimeout fires correctly. Also extended the fix to the stream-json  path added in #139 which had the same issue. Queue chain reset and timer cleanup carried over as-is.

TerrysPOV:
Moved to #142

PR #76, opened 2026-04-05
fix: dual-session fallback to prevent thinking block signature corruption

Nix-molt:
## Summary

Alternative to #26 — addresses the same root cause (issue #18) but preserves session continuity during extended rate-limit windows.

Instead of stripping `--resume` from fallback calls (which makes every fallback call a one-shot with no memory), this PR maintains a **dedicated fallback session** alongside the primary:

- **Separate sessions per provider:** The primary and fallback each have their own session file (`session.json` / `session-fallback.json`). No thinking block signatures are ever mixed.
- **Context handoff:** When the fallback session is first created, the agent is instructed to read recent log files to understand what the primary session was working on. Subsequent fallback calls resume the fallback session normally with full history.
- **No goldfish mode:** During multi-hour rate-limit windows, the fallback session accumulates context just like the primary would — conversations, tasks, and state carry over between calls.

Also includes the `forwardToTelegram`/`forwardToDiscord` fix from #26 (fall back to `result.stdout` when `result.stderr` is empty).

### How it works

1. Primary call hits rate limit → detected via `RATE_LIMIT_PATTERN`
2. Check for existing fallback session (`session-fallback.json`)
3. If no fallback session exists: create one with a handoff prompt that tells the agent to read recent logs for context
4. If fallback session exists: resume it with `--resume` as usual
5. When primary comes back online, the primary session resumes from where it left off

### What this doesn't do (yet)

- **Bidirectional handoff:** When switching _back_ from fallback to primary, the primary doesn't automatically learn what the fallback did. This could be added later.
- **Fallback session cleanup:** The fallback session persists indefinitely. A TTL or manual clear could be added.

## Test plan

- [ ] Configure primary + fallback with different providers
- [ ] Trigger rate limit on primary — verify fallback creates its own session
- [ ] Send multiple messages during rate limit — verify fallback session resumes with context
- [ ] Wait for primary to recover — verify primary session resumes independently
- [ ] Verify Telegram/Discord error messages show actual error text

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -502,7 +502,7 @@ export async function start(args: string[] = []) {
     if (!telegramSend || currentSettings.telegram.allowedUserIds.length === 0) return;
     const text = result.exitCode === 0
       ? `${label ? `[${label}]\n` : ""}${result.stdout || "(empty)"}`
-      : `${label ? `[${label}] ` : ""}error (exit ${result.exitCode}): ${result.stderr || "Unknown"}`;
+      : `${label ? `[${label}] ` : ""}error (exit ${result.exitCode}): ${result.stderr || result.stdout || "Unknown"}`;
     for (const userId of currentSettings.telegram.allowedUserIds) {
       telegramSend(userId, text).catch((err) =>
         console.error(`[Telegram] Failed to forward to ${userId}: ${err}`)
@@ -514,7 +514,7 @@ export async function start(args: string[] = []) {
     if (!discordSendToUser || currentSettings.discord.allowedUserIds.length === 0) return;
     const text = result.exitCode === 0
       ? `${label ? `[${label}]\n` : ""}${result.stdout || "(empty)"}`
-      : `${label ? `[${label}] ` : ""}error (exit ${result.exitCode}): ${result.stderr || "Unknown"}`;
+      : `${label ? `[${label}] ` : ""}error (exit ${result.exitCode}): ${result.stderr || result.stdout || "Unknown"}`;
     for (const userId of currentSettings.discord.allowedUserIds) {
       discordSendToUser(userId, text).catch((err) =>
         console.error(`[Discord] Failed to forward to ${userId}: ${err}`)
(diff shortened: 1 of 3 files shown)


TerrysPOV (changes requested):
This is directionally better than the one-shot fallback approach, but I found one blocking correctness issue and one follow-up problem before merge:

  1. The reverse-handoff path resets the fallback session too early. Right now when the primary comes back, the code prepends a “read recent logs and catch up” handoff to the primary prompt and then immediately deletes session-fallback.json before the primary call has actually succeeded. If that resumed primary call fails, the fallback context is gone even though the system may still need it. The fallback session should only be cleared after a successful primary handoff.
  2. The bundled forwardToTelegram / forwardToDiscord change is only a partial error-message fix. Falling back to raw stdout is better than Unknown, but it still won’t properly extract Claude CLI’s structured JSON error payloads. If this stays in the PR, I’d prefer it to use the same parsed error-detail approach we discussed on the later messaging/error PRs.

  I’d suggest:

  - keep the separate fallback session approach
  - move resetFallbackSession() so it only happens after a successful primary resume/handoff
  - consider narrowing or upgrading the bundled forwarding fix so it properly extracts error detail from JSON stdout, not just stderr || stdout
  - add automated coverage for:
      - first fallback activation creates separate fallback session
      - multiple fallback turns preserve fallback context
      - returning to primary succeeds and only then clears fallback session
      - returning to primary fails and fallback session is preserved
      - forwarded auth/quota errors show useful text
      - close PR26 in favour of this PR

I wasn’t able to verify a clean bunx tsc --noEmit on the branch as submitted because of the separate existing src/config.ts parseSettings signature issue.

squirblej:
Picked this up in #143. Summary of what changed vs the original PR:

The core dual-session concept is intact, but the implementation is adapted to the current codebase (which has changed significantly since this was submitted — `execClaude` now uses `runClaudeStream` unconditionally, sessions are agent-scoped, etc.):

- Fallback sessions are stored as `session_fallback.json` alongside `session.json` (per-agent, same scoping pattern)
- The fallback block rebuilds args from scratch rather than patching the primary args, so the `--resume <primary-session-id>` never leaks through
- No deferred reset — fallback session stays alive across rate-limit events for context continuity
- `extractErrorDetail` is adapted from the closed #111 with a test file

TerrysPOV:
Moved to #143

PR #79, opened 2026-04-07
feat(telegram): add inline button support

Nix-molt:
## Summary

Adds support for sending Telegram inline keyboard buttons via a directive syntax in Claude's response text. Button taps are converted to text and routed back to Claude as a new message — the same way a typed reply would be.

### Syntax

```
[buttons: Yes | No]
[buttons: Accept::confirm | Decline::deny]
```

Each `[buttons:]` directive becomes one row of buttons. Use `Label::data` for custom callback data (defaults to the label). Callback data is namespaced with `kcc_` to avoid colliding with the existing secretary pattern (`sec_yes/no_*`).

### Flow

1. Claude includes `[buttons: ...]` in its response
2. The directive is stripped from visible text; message is sent with `InlineKeyboardMarkup`
3. User taps a button → the label is sent back to Claude as `[Telegram button] <label>` in a new message
4. Claude handles it like any other message, with full session history for context

This is architecturally consistent with the existing `[react:<emoji>]` and `[send-file:<path>]` directives — the tap is just text arriving in the next turn. Agents are prompted about the syntax via SOUL.md (same pattern as `[react:]`).

### New code

- `InlineKeyboardButton` interface
- `BUTTON_CALLBACK_PREFIX = "kcc_"` constant
- `extractButtonDirectives(text)` — parses `[buttons:]` directives
- `sendMessageWithKeyboard(token, chatId, text, buttonRows, threadId?)` — sends with `reply_markup`
- `handleCallbackQuery` extended to handle `kcc_*` button presses and route them back to Claude

Existing `[react:]`, `[send-file:]`, and secretary (`sec_yes/no_*`) flows are unchanged.

## Test plan

- [ ] Send a message that includes `[buttons: Option A | Option B]` — verify buttons appear in Telegram
- [ ] Tap a button — verify Claude receives `[Telegram button] Option A` and responds
- [ ] Verify `[react:]` and `[send-file:]` directives still work alongside buttons
- [ ] Verify secretary `sec_yes/no_*` callback pattern still works
- [ ] Tap a button from an unauthorized user — verify it is silently ignored

### DIFF
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -151,6 +151,11 @@ interface TelegramCallbackQuery {
   data?: string;
 }
 
+interface InlineKeyboardButton {
+  text: string;
+  callback_data: string;
+}
+
 interface TelegramUpdate {
   update_id: number;
   message?: TelegramMessage;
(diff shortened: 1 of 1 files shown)


Nix-molt:
Closing — PR should go against the customized fork at Philipp258/assistants instead.

PR #83, opened 2026-04-09
feat: support integrate Slack bot

hans00:
as title

### DIFF
--- a/commands/slack.md
+++ b/commands/slack.md
@@ -1,0 +1,74 @@
+# Slack Bot — claudeclaw
+
+Start a standalone Slack bot process (Socket Mode).
+
+## Usage
+
+```
+bun run src/index.ts slack
+```
+
+Or via the daemon (`bun run src/index.ts start`), which auto-starts the bot when `slack.botToken` and `slack.appToken` are configured.
+
+## Configuration
+
+Add to `.claude/claudeclaw/settings.json`:
+
+```json
+{
+  "slack": {
+    "botToken": "xoxb-...",
+    "appToken": "xapp-...",
+    "allowedUserIds": [],
+    "listenChannels": []
+  }
+}
+```
+
+| Field | Description |
+|---|---|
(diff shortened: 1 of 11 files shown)


saravmajestic:
👍 useful for me

TerrysPOV (changes requested):
This is a substantial addition, but I found two blocking correctness issues in the current PR tip, including the later follow-up commits:

  1. Assistant-thread mode is currently tracked at channel scope instead of thread scope. assistant_thread_started / assistant_thread_context_changed only record channel_id, and the message gate later checks assistantThreadChannels.has(channelId). That means once one Assistant thread is opened in a channel, the bot will start replying to unrelated non-mention messages across the whole channel. This needs to stay scoped to the specific assistant thread, not the entire channel.
  2. The new [read_channel:...] flow is incomplete. The code fetches channel history, saves it to disk, and sends a second runUserMessage() telling Claude to read/summarize it, but the output of that second run is never sent back to Slack. As written, a directive-only read_channel response can leave the user with no summary and sometimes no visible reply at all.

  I’d suggest:

  - track Assistant-thread allowlisting by (channel_id, thread_ts) rather than channel_id alone
  - make the read_channel flow actually surface the second-pass result back to Slack, or redesign it so the first pass waits for the read/summarize round-trip before replying
  - rebase or restack this branch on top of the existing parseSettings signature fix so it can be typechecked cleanly
  - add automated coverage for:
      - normal channel message still requiring @mention
      - assistant thread reply working without @mention
      - opening one assistant thread not unlocking the whole channel
      - [read_channel:...] producing a user-visible summary/response

I was able to run bun install in the PR worktree, but bunx tsc --noEmit still fails there on the separate existing src/config.ts parseSettings signature issue, so I could not verify the branch as fully type-clean.

squirblej:
Picked this up in #151. The PR branch was too far behind to rebase cleanly — the config/runner/discord changes would have regressed several recently merged PRs. Instead ported only the Slack-specific content onto current master and fixed both blockers (assistant-thread scope, read_channel result dropped) plus the secondary issues from the Opus review.

TerrysPOV:
Closed as now in 151

PR #85, opened 2026-04-09
feat: add multi-session chat UI with session browser and agent picker

saravmajestic:
Adds a session sidebar and agent selector to the web UI chat tab, enabling users to browse, resume, and create sessions across all channels.

Server-side:
- New GET /api/sessions — lists all active sessions from session.json, sessions.json (thread sessions), and JSONL files in ~/.claude/projects/
- New GET /api/sessions/:id/messages — reads JSONL session files, extracts user/assistant messages with pagination (limit + offset params)
- New GET /api/agents — lists available agents from .claude/agents/ dirs (supports both workspace layout agent/<name>/agent.md and legacy flat files)

UI changes:
- Session sidebar (left panel) with session list showing agent name, channel badge (global/slack/whatsapp), message preview, timestamp, turns
- Click session to load its message history
- "Load older" button at top for paginated history (loads 10 at a time)
- Scroll position preserved when loading older messages
- "+ New" button opens clean chat with agent dropdown above input
- Agent dropdown defaults to Mike, hides after first message sent
- Mobile responsive — sidebar collapses on small screens

JSONL parsing:
- Reads Claude Code session files from ~/.claude/projects/<project>/*.jsonl
- Extracts user messages (strips ClaudeClaw prefixes: timestamps, channel info, Slack directives)
- Extracts assistant text (skips thinking blocks and tool_use)
- Supports offset=-1 for "last N messages" loading

### DIFF
--- a/src/ui/page/script.ts
+++ b/src/ui/page/script.ts
@@ -943,6 +943,11 @@ export const pageScript = String.raw`    const $ = (id) => document.getElementBy
     const chatSend = $("chat-send");
 
     var CHAT_STORAGE_KEY = "claudeclaw.chat.history";
+    var activeSessionId = null;
+    var activeAgent = null;
+    var messagesOffset = 0;
+    var totalMessages = 0;
+    var MESSAGES_PER_PAGE = 10;
     let chatBusy = false;
     let chatAbortController = null;
     let chatElapsedTimer = null;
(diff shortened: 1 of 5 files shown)


TerrysPOV (changes requested):
This is a useful UI direction, but I found two blocking product/correctness issues in the current implementation:

  1. The new session browser does not actually resume or target the selected session. The UI stores activeSessionId, but sendChat() still posts only { message }, the /api/chat endpoint only accepts message, and the server still always routes that through the default web chat path. So selecting a historical session only changes the visible transcript; the next message does not go into that selected session.
  2. The new agent picker is also cosmetic right now. The UI updates activeAgent, but that value is never sent to the backend and there is no server-side path that uses it. So the PR presents agent selection as a feature without implementing the backend behavior.

I also couldn’t verify a clean typecheck on the branch as submitted because it still carries stale-base signature issues (parseSettings(...) and json(..., status)), so this branch should be rebased before final validation anyway.

  I’d suggest:

  - wire the selected sessionId through /api/chat and into the actual runner/session selection path, or remove the “resume/browse sessions” implication for now
  - wire the selected agent through the server and into the backend prompt/agent routing path, or remove the picker until that exists
  - rebase onto current master / the existing signature-fix branches before final validation
  - add coverage for:
      - selecting an existing session and sending a follow-up into that same session
      - starting a new session with a selected agent
      - no mismatch between displayed transcript and actual backend target session

bun install succeeded in /private/tmp/claudeclaw-pr85-review. bunx tsc --noEmit fails with the stale parseSettings / json() signature issues above. I did not find automated tests covering the new session-browser or agent-picker flows. Please consider adding these in.

squirblej:
I did a detailed review of this PR against current master. Sharing findings in case they're useful for whoever picks this up — I don't have time to implement the fixes right now.

### The core wiring problem

The fundamental blocker (beyond what @TerrysPOV flagged) is that `streamClaude`/`streamUserMessage` in runner.ts don't accept `agentName` or `threadId` at all — those parameters were only added to `execClaude`. Wiring session/agent targeting requires extending the streaming path to do its own session lookup, session creation on first turn, and spawn-cwd handling. That's the main piece of work.

### Additional issues in `services/sessions.ts`

**Wrong agent directory.** `listAgents()` reads `<cwd>/.claude/agents/` and looks for `agent.md` files with `name:`/`description:` frontmatter. ClaudeClaw agents live at `<cwd>/agents/<name>/` (see `getAgentsDir()` in config.ts). There's no `agent.md` convention — an agent is just a directory. The result is `listAgents()` always returns only the hardcoded `mike` entry.

**All sessions labeled `agent: "mike"`.** `listSessions()` never reads the per-agent `session_{name}.json` files, so every session entry (global, thread, orphan) is tagged as `mike` regardless of which agent owns it.

**Fabricated channel taxonomy.** `threadId.startsWith("slk:") ? "slack" : "whatsapp"` — neither `slk:` prefix nor a WhatsApp integration exists in this codebase. Discord thread IDs are raw 17-19 digit snowflakes.

**`getProjectDir()` sanitizer is incomplete.** `replace(/\//g, "-")` only handles slashes; Claude Code's actual sanitizer also handles `.` in paths. On a project path like `/home/user/my.project`, this produces the wrong JSONL directory.

**Slow double-parse per session.** `listSessions()` calls `readSessionMessages()` twice per entry (first + last message), which does a full JSONL file read each time. With dozens of sessions this is O(N×file_size) on every sidebar refresh.

**Multi-line message truncation.** `readSessionMessages()` strips prefixes then does `text.split("\n").pop()` — keeping only the last line. Any multi-line user message (code, errors, pasted content) is silently truncated. Should strip the prefix block and keep the rest as-is.

### Other issues

**Path traversal.** `/api/sessions/:id/messages` extracts the session ID by string replace with no validation before building the file path. Should require UUID shape (`/^[a-f0-9-]{36}$/i`).

**TypeScript error.** `json({ error: String(err) }, 500)` — the `json()` helper in `src/ui/http.ts` only accepts one argument. Three call sites in the new endpoints.

**`json(x, 500)` signature issue is separate from the stale-base issue** — it exists in the PR's own code.

**Dead UI element.** The `#agent-selector` sidebar dropdown is hidden by default and never shown — only `#inline-agent-selector` is toggled. Can be removed.

**`setActiveTab` monkey-patch** (`var origSetActiveTab = setActiveTab; setActiveTab = function(...)`) is fragile and will break silently if the surrounding script is ever refactored to `const`.

### Out of scope for v1

Thread sessions (Discord) should be read-only in the web sidebar — replying from web would inject into a Discord thread without the Discord users seeing the input, which is confusing. Worth disabling the input or showing a read-only badge for those entries.

---

Hope this saves someone some time.

TerrysPOV:
@squirblej thanks for this. Very useful. I've picked this up.

TerrysPOV:
Superseded by #153, which ports this work with all issues from squirblej's review applied. Attribution to both @saravmajestic and @squirblej included in the commit and PR.

PR #87, opened 2026-04-10
feat: daemon-level plugin system with OpenClaw-compatible API

Amineelfarssi:
## Summary

- Adds a generic plugin system for ClaudeClaw daemon-level lifecycle events
- Plugins use the same API shape as OpenClaw plugins (`api.on()`, `api.registerService()`, `api.registerCommand()`), so existing OpenClaw plugins like **claude-mem** work without modification
- Core only knows about events and the plugin API — it doesn't know any specific plugin exists
- Zero overhead when no plugins are configured (early returns on empty handler maps)

## Architecture

```
┌─────────────────────────────────────────────┐
│  ClaudeClaw Daemon                          │
│  ┌─ daemon plugin events ─────────────────┐ │
│  │ before_agent_start                     │ │
│  │ before_prompt_build → inject context   │ │
│  │                                        │ │
│  │   ┌─ claude -p ──────────────────────┐ │ │
│  │   │  CC plugin hooks fire here       │ │ │
│  │   │  SessionStart, PostToolUse, etc. │ │ │
│  │   └──────────────────────────────────┘ │ │
│  │                                        │ │
│  │ agent_end → summarize                  │ │
│  └────────────────────────────────────────┘ │
└─────────────────────────────────────────────┘
```

Daemon plugins fire **around** Claude Code invocations (session lifecycle, prompt injection, channel routing). Claude Code's native plugins fire **inside** them (tool-level hooks). Both coexist without conflict.

## Changes

| File | Change |
|---|---|
| `src/plugins.ts` | **NEW** — PluginManager, PluginApi interface, event types, plugin loader/resolver (362 lines) |
| `src/runner.ts` | Event emission at lifecycle points in `execClaude` and `streamClaude` (+56 lines) |
| `src/config.ts` | `plugins` field in Settings interface + parser (+4 lines) |
| `src/commands/start.ts` | Plugin init/shutdown, channel sender wiring (+28 lines) |

## Settings format

```json
"plugins": {
  "claude-mem": {
    "enabled": true,
    "source": "openclaw",
    "config": { "workerPort": 37777, "project": "myproject" }
  }
}
```

## Lifecycle events

`gateway_start`, `session_start`, `before_agent_start`, `before_prompt_build`, `tool_result_persist`, `agent_end`, `session_end`, `message_received`, `after_compaction`

## Test plan

- [x] Plugin loads and initializes on daemon start
- [x] `before_agent_start` → session init in claude-mem worker
- [x] `before_prompt_build` → memory context injected into system prompt
- [x] `tool_result_persist` → observations captured (streaming: per-tool, non-streaming: per-run)
- [x] `agent_end` → summarization triggered
- [x] `after_compaction` → session preserved
- [x] Zero overhead with no plugins configured
- [x] Tested with claude-mem OpenClaw plugin (v12.1.0) end-to-end

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -10,6 +10,7 @@ import { initConfig, loadSettings, reloadSettings, resolvePrompt, type Heartbeat
 import { getDayAndMinuteAtOffset } from "../timezone";
 import { startWebUi, type WebServerHandle } from "../web";
 import type { Job } from "../jobs";
+import { PluginManager, setPluginManager } from "../plugins";
 
 const CLAUDE_DIR = join(process.cwd(), ".claude");
 const HEARTBEAT_DIR = join(CLAUDE_DIR, "claudeclaw");
@@ -319,7 +320,18 @@ export async function start(args: string[] = []) {
   let web: WebServerHandle | null = null;
   let discordStopGateway: (() => void) | null = null;
 
+  // ── Plugin system ──
+  const pluginManager = new PluginManager(process.cwd());
+  if (Object.keys(settings.plugins).length > 0) {
+    await pluginManager.loadAll(settings.plugins);
+    await pluginManager.startServices();
+    await pluginManager.emit("gateway_start", {}, { workspaceDir: process.cwd() });
+    setPluginManager(pluginManager);
+  }
+
   async function shutdown() {
+    await pluginManager.stopServices();
+    setPluginManager(null);
     if (discordStopGateway) discordStopGateway();
     if (web) web.stop();
     await teardownStatusline();
@@ -394,6 +406,22 @@ export async function start(args: string[] = []) {
 
   await initDiscord(currentSettings.discord.token);
   if (!discordToken) console.log("  Discord: not configured");
+
+  // Wire channel senders into plugin runtime
+  if (pluginManager.hasPlugins) {
+    pluginManager.setChannelSenders({
+      telegram: {
+        sendMessageTelegram: telegramSend
+          ? (chatId: number, text: string) => telegramSend!(chatId, text)
+          : () => Promise.resolve(),
+      },
+      discord: {
+        sendMessageDiscord: discordSendToUser
+          ? (userId: string, text: string) => discordSendToUser!(userId, text)
+          : () => Promise.resolve(),
+      },
+    });
+  }
 
   function isAddrInUse(err: unknown): boolean {
     if (!err || typeof err !== "object") return false;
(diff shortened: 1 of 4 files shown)


TerrysPOV (changes requested):
This is an interesting direction, and I checked the OpenClaw side before finalizing this review. OpenClaw itself appears to define a broader plugin hook surface than the subset that is actually wired into execution in some versions, so I would not block this PR just because some lifecycle event names exist in the API without all of them firing yet. In other words, that part does seem consistent with the compatibility pattern it is trying to mirror.

The blocking issue I do see is with tool_result_persist. Right now this PR uses that event name, but the data being sent through it does not look like an actual tool result. In the non-streaming path it fires once for the whole Claude run, using the daemon call label as toolName, the whole prompt as params, and final assistant stdout as the message content. In the streaming path it fires when a tool is invoked (tool_use), not when the tool result is completed. So a plugin expecting real tool-result observations would receive synthetic run-level data instead. That feels like a real compatibility/correctness problem, not just an API-shape difference.

  Two additional follow-ups:

  - The PR description currently reads more strongly than the implementation. Since not every declared lifecycle event is actually emitted in ClaudeClaw yet, I’d suggest describing this as an OpenClaw-compatible plugin API surface with a currently implemented subset of active hooks, rather than implying the full listed event set is live here today.
  - Plugin config is startup-only right now. Plugins load at daemon start, but the existing hot-reload loop does not reload settings.plugins, so enabling, disabling, or reconfiguring plugins still requires a daemon restart. That may be acceptable, but it should be explicit.

I also couldn’t verify a clean typecheck on the submitted branch because it still carries the separate existing src/config.ts parseSettings signature issue. Please rebase this branch onto a base that already includes that fix before final validation.

  I’d suggest:

  - make tool_result_persist represent actual tool results, or rename/scope it so plugin authors are not misled about what event they’re receiving
  - narrow the PR description so it clearly separates “compatible API surface” from “hooks currently emitted by ClaudeClaw”
  - document or implement the plugin reload behavior explicitly
  - add automated coverage for emitted event payloads, especially the difference between non-streaming runs and streaming tool-use flows

squirblej:
Hi — I started working on this and opened #152 addressing the review feedback (correct tool_result_persist pairs, trimmed DaemonEvent to wired-only events, added unloadAll and session_start emit). I then noticed that you'd already merged your own implementation directly to master as a421b4b before I opened the PR, so I closed #152.

Should this PR be closed now that the feature is in master? Happy to either close it here or add anything from my implementation that wasn't in your version (unloadAll, session_start emit) as a follow-up if useful.

TerrysPOV:
Closed per comment above

PR #88, opened 2026-04-10
fix: null-safe mentions check + sessionTimeoutMs config parsing

justmaker:
## Bug 1: `message.mentions` TypeError crash

Discord gateway `MESSAGE_CREATE` events sometimes lack the `mentions` array. This causes a `TypeError: undefined is not an object (evaluating 'message.mentions.some')` crash in `guildTriggerReason()`, leading to a systemd restart loop.

**Fix:** Optional chaining `message.mentions?.some()`

## Bug 2: `sessionTimeoutMs` setting ignored

The `sessionTimeoutMs` field in `settings.json` is never parsed by `parseSettings()`, so `runner.ts`'s `(settings as any).sessionTimeoutMs` always returns `undefined`, falling back to the hardcoded 5-minute timeout.

**Fix:**
- Add `sessionTimeoutMs: number` to `Settings` interface
- Parse it in `parseSettings()` (default: 0 = use CLAUDE_TIMEOUT_MS)
- Replace `(settings as any).sessionTimeoutMs` with typed `settings.sessionTimeoutMs` in `runner.ts`

## Files changed
- `src/commands/discord.ts` — optional chaining on mentions
- `src/config.ts` — Settings interface + DEFAULT_SETTINGS + parseSettings
- `src/runner.ts` — typed sessionTimeoutMs access

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 .claude
 node_modules
 .codemachine
-AGENTS.md
+AGENTS.md__pycache__/
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -264,7 +264,7 @@ function guildTriggerReason(message: DiscordMessage): string | null {
   if (botUserId && message.referenced_message?.author?.id === botUserId) return "reply_to_bot";
 
   // Mention via mentions array
-  if (botUserId && message.mentions.some((m) => m.id === botUserId)) return "mention";
+  if (botUserId && message.mentions?.some((m) => m.id === botUserId)) return "mention";
 
   // Mention in content (fallback)
   if (botUserId && message.content.includes(`<@${botUserId}>`)) return "mention_in_content";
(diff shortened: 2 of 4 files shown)


TerrysPOV (changes requested):
I think there is a real fix here, but the PR should be narrowed before merge.

  1. The Discord message.mentions?.some(...) change looks valid and does address a real crash path if mentions is missing on some gateway payloads.
  2. The sessionTimeoutMs part appears to already be fixed on current master. Current master already parses sessionTimeoutMs in config.ts and uses typed settings.sessionTimeoutMs in runner.ts, so that part of the PR description is stale relative to the current base.
  3. There is also an unrelated .gitignore change in this branch replacing AGENTS.md with AGENTS.md__pycache__/, which doesn’t seem related to the bugfix and looks incorrect.

  I’d suggest:

  - keep the Discord null-safe mentions fix
  - drop the stale sessionTimeoutMs framing from the PR description
  - remove the unrelated .gitignore change
  - rebase onto current master / the existing parseSettings fix branch before final validation

I wasn’t able to verify a clean bunx tsc --noEmit on the branch as submitted because it still carries the separate existing src/config.ts parseSettings signature issue unrelated to this PR.

squirblej:
Addressed in #141 — narrowed to just the `mentions?.some()` fix as suggested.

Dropped the `sessionTimeoutMs` changes (already on master via #129) and the `.gitignore` entry (had a missing newline producing `AGENTS.md__pycache__/`). Single-file, single-character change, `bunx tsc --noEmit` clean.

TerrysPOV:
Moved to #141

PR #89, opened 2026-04-10
feat: add listenGuilds for guild-wide message listening

justmaker:
## Problem

When using `listenChannels` to listen to a specific Discord channel, threads created in that channel rely on an in-memory `knownThreads` map to track parent-child relationships. This map is volatile — after a restart, threads not persisted in `sessions.json` become unresponsive because the bot doesn't recognize them as belonging to a listened channel.

## Solution

Add `discord.listenGuilds`: an array of guild IDs where the bot responds to **all messages** in any channel or thread, without requiring mention or per-channel setup.

```json
{
  "discord": {
    "listenGuilds": ["123456789012345678"]
  }
}
```

This is checked in `guildTriggerReason()` after `listenChannels` — if the message's `guild_id` matches, the bot responds immediately without needing the thread's parent to be tracked.

## Files changed
- `src/config.ts` — `DiscordConfig.listenGuilds`, DEFAULT_SETTINGS, parseSettings
- `src/commands/discord.ts` — `listen_guild` trigger reason

## Backward compatible
- `listenGuilds` defaults to `[]` — no behavior change unless configured
- Existing `listenChannels` behavior is unchanged

### DIFF
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -273,6 +273,9 @@ function guildTriggerReason(message: DiscordMessage): string | null {
   const config = getSettings().discord;
   if (config.listenChannels.includes(message.channel_id)) return "listen_channel";
 
+  // Listen guild (respond to all messages in any channel/thread of this guild)
+  if (message.guild_id && config.listenGuilds?.includes(message.guild_id)) return "listen_guild";
+
   // Thread whose parent channel is a listen channel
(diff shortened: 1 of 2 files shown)


TerrysPOV:
Thanks @justmaker — this has been picked up, rebased cleanly onto current master, and the startup logging has been extended to cover `listenGuilds` too. A new PR is open at #148. Closing this one in favour of it.

PR #92, opened 2026-04-14
Add current clock to heartbeat and job prompts

christian-drescher:
This PR adds current clock to heartbeat and job prompts. The agent was previously not aware of the current time when running heartbeats or jobs.

**Previously:**
The date in the agent's context is a snapshot from session start. If a session runs long enough, it will drift as the currentDate in the system reminder gets stale relative to when it was loaded. User messages, e.g., from Telegram, help to peace together the current time.

**Proposed Change:**
Inject current clock to heartbeat and job prompt.

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -7,7 +7,7 @@ import { cronMatches, nextCronMatch } from "../cron";
 import { clearJobSchedule, loadJobs } from "../jobs";
 import { writePidFile, cleanupPidFile, checkExistingDaemon } from "../pid";
 import { initConfig, loadSettings, reloadSettings, resolvePrompt, type HeartbeatConfig, type Settings } from "../config";
-import { getDayAndMinuteAtOffset } from "../timezone";
+import { getDayAndMinuteAtOffset, buildClockPromptPrefix } from "../timezone";
 import { startWebUi, type WebServerHandle } from "../web";
 import type { Job } from "../jobs";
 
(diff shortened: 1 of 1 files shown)


TerrysPOV (changes requested):
Nice one! Simple and really useful. 

I don’t see a blocking issue in this change. It looks like the right fix for the gap between runUserMessage() and the non-user run() paths: interactive/user-driven messages already get a fresh clock prefix, while heartbeat and cron jobs did not. Adding the current clock before heartbeat and job prompts is consistent with the existing prompt model and addresses the stale-date drift this PR describes.

I wasn’t able to verify a clean bunx tsc --noEmit on the branch as submitted because it still carries the separate existing src/config.ts parseSettings signature issue unrelated to this PR. Please rebase this branch onto a base that already includes that fix before final validation.

TerrysPOV (changes requested):
Review scope: full current `#92` head against current `master`, including the latest push after the earlier review comment.

  ## High Priority 🟠

  1. **Missing required version bumps for shipped runtime changes**
     - **Problem:** this PR changes shipped code under `src/commands/start.ts` but does not update `.claude-plugin/plugin.json` or `.claude-plugin/marketplace.json`.
     - **Impact:** under the current version-guard workflow rules, this should require both version bumps.

  I re-checked the full current PR tip against current `master` and did not find a further functional issue in the clock-prefix change itself.

christian-drescher:
Hey! The plugin-version-guard and marketplace-version-guard checks require every PR to bump versions, but when multiple PRs are in flight they're all racing to bump from the same base. The second one to merge will conflict (or silently produce a duplicate version).

Would it make sense to either:

- Run the version bump automatically on merge (e.g. in a post-merge CI step), or
- The project maintainers decide to version bump at their discretion when it is meaningful rather than per-PR?

TerrysPOV:
OK i hear you. The last thing I want is unnecessary friction for you guys who are committed and contributing so readily. I've removed these checks for now while I consider the best approach.

TerrysPOV:
Thanks again @christian-drescher. I have combined this fix into #145 so we can land the accepted clock-prefix work together with the other small stability fixes while avoiding version-bump churn across multiple open PRs. I preserved attribution for your contribution in the combined branch.

PR #93, opened 2026-04-15
Adds notify frontmatter support to job creation and editing

christian-drescher:
Adds the notify option to the job management commands so users can control whether job output is forwarded to Telegram when creating, listing, or editing jobs.

### DIFF
--- a/commands/config.md
+++ b/commands/config.md
@@ -40,7 +40,7 @@ Parse `$ARGUMENTS` to identify what the user wants. If no arguments are given, s
    - Enabled: yes/no
    - Address: host:port
 
-3. Also list any cron jobs from `.claude/claudeclaw/jobs/` with their name and schedule.
+3. Also list any cron jobs from `.claude/claudeclaw/jobs/` with their name, schedule, and notify mode.
 4. Remind the user that changes are hot-reloaded every 30s — no daemon restart needed.
 
 ### `heartbeat on` / `heartbeat off` / `heartbeat enable` / `heartbeat disable`
--- a/commands/config.md
+++ b/commands/config.md
@@ -40,7 +40,7 @@ Parse `$ARGUMENTS` to identify what the user wants. If no arguments are given, s
    - Enabled: yes/no
    - Address: host:port
 
-3. Also list any cron jobs from `.claude/claudeclaw/jobs/` with their name, schedule, and notify mode.
+3. Also list any cron jobs from the configured jobs directory (see `jobsDir` in settings, default: `.claude/claudeclaw/jobs/`) with their name, schedule, and notify mode.
 4. Remind the user that changes are hot-reloaded every 30s — no daemon restart needed.
 
 ### `heartbeat on` / `heartbeat off` / `heartbeat enable` / `heartbeat disable`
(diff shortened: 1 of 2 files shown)


TerrysPOV (changes requested):
This is useful overall. The docs are now much closer to the existing job behaviour, especially around notify, which is already supported by the runtime.

One issue I’d fix before merge: the new edit flow asks AskUserQuestion with four options (Schedule, Prompt, Notify, All). That doesn’t match the current interaction constraint for these questions, which is 2–3 mutually exclusive choices. So I’d suggest restructuring that part of the doc into a sequence of smaller questions, or reducing the initial choice set so it fits the supported UI pattern.

Other than that, this looks like a worthwhile docs update.

christian-drescher:
I have changed the edit flow by removing the "Edit All" option, so every AskUserQuestion call stays within the 2–3 option constraint.

The users can still edit all fields by looping. I have introduced a new/last step before writing changes to file.

- Step 3: Asks with 3 options (Schedule, Prompt, Notify) instead of 4
- Step 4: Applies the chosen change (no more All branch)
- Step 5: Asks "Anything else to change?" (Yes/No) — loops back to step 3 if yes, writes and confirms if no

TerrysPOV (changes requested):
Review scope: full current `#93` head against current `master`, including the new comment describing the edit-flow loop update.

  ## High Priority 🟠

  1. **The docs now hard-code the jobs directory, but runtime supports configurable `jobsDir`** (`commands/jobs.md`, `commands/config.md`)
     - **Problem:** the PR repeatedly instructs users to read and write `.claude/claudeclaw/jobs/` directly.
     - **Impact:** on current `master`, job storage can be redirected via `jobsDir`, so these instructions can point users at the wrong files and make the commands appear broken.
     - **Fix:** describe the default path as the default only, and instruct the command to respect configured `jobsDir` rather than assuming `.claude/claudeclaw/jobs/` unconditionally.

  2. **`notify` is documented as Telegram-only, but runtime applies it to both Telegram and Discord forwarding** (`commands/jobs.md`, current `src/commands/start.ts`)
     - **Problem:** the new wording says “Should this job notify you on Telegram?” and the reference section describes `notify` only in terms of Telegram.
     - **Impact:** this is no longer accurate against the current runtime, where the same `notify` gate controls whether job output is forwarded to both Telegram and Discord.
     - **Fix:** update the wording to describe notification forwarding generically, or explicitly mention both Telegram and Discord.

  3. **Missing required version bumps for shipped command changes**
     - **Problem:** this PR changes shipped files under `commands/` but does not update `.claude-plugin/plugin.json` or `.claude-plugin/marketplace.json`.
     - **Impact:** under the current version-guard workflow rules, this should require both version bumps.

christian-drescher:
The review asks to fix issues that are unrelated to this PR (i.e., the issues existed before).

This PR adds a config option that exists in the runtime, but was not exposed to the user. However, the review relates to hard-coded paths in the file that were not introduced in this PR.

I have pushed the requested "changes", anyhow.

Note that the PR does not bump versions, because this will conflict with other PRs.

TerrysPOV (approved):
Thanks for this. This is good to merge and will be cherry picked in wiht the next batch going later today.

TerrysPOV:
Thanks @christian-drescher for the contribution. I have cherry-picked these changes into #154 as part of the bulk merge PR, preserving the feature/fix commits and leaving out the upstream merge commits, so I am closing this source PR now. Appreciate the work here.

Issue #51, opened 2026-03-20
Superfluous addition of prompt templates with every invocation


christian-drescher:
Reviewing `runner.ts`, the function `execClaude` in particular, I noticed that through `loadPrompts()` the contents of the *empty template files* `IDENTITY.md`, `SOUL.md`, and `USER.md` are always appended to Claude's system prompt in addition to any existing `CLAUDE.md`.

A comment in the code states that prompt files must be passed on every invocation because appendices to system prompts do not persist across `--resume` (session resumption).

Still, I find the addition of mentioned prompt files superfluous.

1. Is `CLAUDE.md` not loaded by Claude Code automatically? If true, then adding the contents of `CLAUDE.md` to the system prompt is introducing duplicates.

2. Beyond bootstrapping `CLAUDE.md` (if nonexistent), adding `IDENTITY.md`, `SOUL.md`, and `USER.md` to the system prompt **every time** serves no purpose as those are largely empty templates.

I think this is a bug.

TerrysPOV:
I don't think this is a bug, I think it is a function of how claude p --resume works. In this mode (which is the only current claude code mode that can support a daemon/gateway) claude.md and other "system" files (SOUL.md, USER.md, IDENTITY.md, all Skills, plugins and MCPs) are loaded with every invocation. This cannot be changed as it is a inherent to how claude code works. Anthropic would need to make some not insignificant changes to claude code in order to support this not being the case.

I started with a fairly minimal config and was loading 32k token for every message I sent. I've had to scale this back and I'm now at 15k per message. i'll be scaling back even further. The problem with this approach is that it means less skills, less MCPs and no plugins (as these are a major killer) and this of course makes it less useful. Instead of using MCPs I'm using CLIs as much as possible as this saves token too. I'm quickly concluding that this maybe a good general agent, but I won't be delegating context heavy workloads (i.e. coding) to ClaudeClaw. 

P.S. The ClaudeClaw plugin itself consumes 10k of tokens for every message. I've removed it completely and am running it standalone now and starting/stopping it manually from my terminal.

christian-drescher:
I understand that the agent's description etc. has to be loaded with every message as they contain useful information. My issue is with the inclusion of **template files** that hold no useful information that the agent can act upon (because there are templates).

TerrysPOV:
[...]

PR #94, opened 2026-04-15
fix: Superfluous addition of prompt templates with every invocation

christian-drescher:
This PR addresses some points raised in issue #51, namely the contents of `CLAUDE.md` (that is created from the *template* documents `IDENTITY.md`, `USER.md`, `SOUL.md`) **AND** the *templates* themselves, all are added twice in every system prompt.

**IMPORTANT:** According to the documentation of Claude Code, the contents of `CLAUDE.md` are added automatically as the first user message to every session. Hence, we might have contents of `CLAUDE.md` in every session twice even after accepting this PR (trice before this PR). This issue remains untouched by this PR. *I propose renaming `CLAUDE.md` to something else.*

**Technical Description**
`ensureProjectClaudeMd()` already embeds the prompt files (`IDENTITY.md`, `USER.md`, `SOUL.md`) into the project-root `CLAUDE.md` via managed blocks, and it runs before every `execClaude()` call. Despite this, both functions were independently calling `loadPrompts()` and appending the raw prompt content to `--append-system-prompt` **in addition to** reading `CLAUDE.md`, resulting in the same template text appearing **twice in every system prompt**. Meanwhile, `CLAUDE.md` is the file that actually evolves over time (user and agent edit it), while the plugin's prompt files `IDENTITY.md`, `USER.md`, `SOUL.md` remain static *templates* unless someone finds out about them.

This PR removes the `loadPrompts()` calls from `execClaude()` and `streamClaude()` so both now read solely from `CLAUDE.md` as the single source of truth for appended system prompt content. `loadPrompts()` itself is retained because it's still used by `ensureProjectClaudeMd()` to generate the managed block during initial `CLAUDE.md` creation. If `CLAUDE.md` can't be read at runtime (e.g. the earlier write failed), the invocation proceeds with just the baseline "You are running inside ClaudeClaw." preamble rather than silently falling back to the raw templates.

### DIFF
--- a/src/commands/send.ts
+++ b/src/commands/send.ts
@@ -1,4 +1,4 @@
-import { runUserMessage } from "../runner";
+import { runUserMessage, ensureProjectClaudeMd } from "../runner";
 import { getSession } from "../sessions";
 import { loadSettings, initConfig } from "../config";
 
@@ -14,6 +14,7 @@ export async function send(args: string[]) {
 
   await initConfig();
   await loadSettings();
+  await ensureProjectClaudeMd();
 
   const session = await getSession();
   if (!session) {
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -393,23 +393,20 @@ async function execClaude(name: string, prompt: string, threadId?: string): Prom
     args.push("--resume", existing.sessionId);
   }
 
-  // Build the appended system prompt: prompt files + directory scoping
+  // Build the appended system prompt: CLAUDE.md + directory scoping
   // This is passed on EVERY invocation (not just new sessions) because
   // --append-system-prompt does not persist across --resume.
-  const promptContent = await loadPrompts();
+  // Prompt files (IDENTITY.md, USER.md, SOUL.md) are already embedded in
+  // CLAUDE.md by ensureProjectClaudeMd(), which runs before every call.
   const appendParts: string[] = [
     "You are running inside ClaudeClaw.",
   ];
-  if (promptContent) appendParts.push(promptContent);
 
-  // Load the project's CLAUDE.md if it exists
-  if (existsSync(PROJECT_CLAUDE_MD)) {
-    try {
-      const claudeMd = await Bun.file(PROJECT_CLAUDE_MD).text();
-      if (claudeMd.trim()) appendParts.push(claudeMd.trim());
-    } catch (e) {
-      console.error(`[${new Date().toLocaleTimeString()}] Failed to read project CLAUDE.md:`, e);
-    }
+  try {
+    const claudeMd = await Bun.file(PROJECT_CLAUDE_MD).text();
+    if (claudeMd.trim()) appendParts.push(claudeMd.trim());
+  } catch (e) {
+    console.error(`[${new Date().toLocaleTimeString()}] Failed to read project CLAUDE.md:`, e);
   }
 
   if (security.level !== "unrestricted") appendParts.push(DIR_SCOPE_PROMPT);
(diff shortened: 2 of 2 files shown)


TerrysPOV (changes requested):
As you know, given we have discussed this, I do think the main idea here is right: if ensureProjectClaudeMd() is the mechanism that materializes the managed prompt templates into project-root CLAUDE.md, then reading both the raw templates and CLAUDE.md on every invocation is redundant and likely does duplicate prompt content.

One follow-up issue before merge, though: this change removes the raw-template fallback, so every invocation path now depends on CLAUDE.md already existing. Most of the main entry points call ensureProjectClaudeMd() first, but claudeclaw send currently does not. send.ts loads settings and calls runUserMessage() directly, so on a repo where CLAUDE.md hasn’t been created yet, this PR would now drop the managed prompt templates for that path instead of falling back to them.

  I’d suggest either:

  - adding ensureProjectClaudeMd() to send.ts, or
  - keeping a narrowly scoped fallback for paths where CLAUDE.md has not been initialized yet

I also couldn’t verify a clean bunx tsc --noEmit on the branch as submitted because it still carries the separate existing src/config.ts parseSettings signature issue unrelated to this PR. Please rebase this branch onto a base that already includes that fix before final validation.

TerrysPOV:
Hi @christian-drescher 

In addition to the above comment, I do think the main idea in PR #94 is right: if ensureProjectClaudeMd() is the mechanism that materialises the managed prompt templates into project-root CLAUDE.md, then reading both the raw templates and CLAUDE.md on every invocation is redundant and likely does duplicate prompt content.

One follow-up issue before that approach is really complete, though: removing the raw-template fallback means every invocation path now depends on CLAUDE.md already existing. Most of the main entry points call ensureProjectClaudeMd() first, but claudeclaw send currently does not. send.ts loads settings and calls runUserMessage() directly, so on a repo where CLAUDE.md has not been created yet, that path would now drop the managed prompt templates instead of falling back to them.

  So my current view is:

  - #94 is directionally the right fix for the duplicate-template part of this issue
  - but it still needs the invocation-path cleanup so CLAUDE.md is guaranteed to exist wherever that assumption is now required

  I’d suggest either:

  - adding ensureProjectClaudeMd() to send.ts, or
  - keeping a narrowly scoped fallback for paths where CLAUDE.md has not been initialised yet

christian-drescher:
I would add `ensureProjectClaudeMd()` to send.ts. A *narrowly scoped fallback* would add a second code path to maintain and be asymmetric with every other entry point.

send.ts didn't have `ensureProjectClaudeMd()` because it is guarded by `getSession()` (*"No active session. Start the daemon first."*), implying the daemon was started, which should have already created `CLAUDE.md` via start.ts.

TerrysPOV (changes requested):
Review scope: full current `#94` head against current `master`, including the updated PR discussion and the latest author reply about `send.ts`.

  ## High Priority 🟠

  1. **`send` still depends on an unenforced `CLAUDE.md` invariant** (`src/commands/send.ts`, `src/runner.ts`)
     - **Problem:** this PR removes the raw-template fallback in `runner.ts` and now assumes project-root `CLAUDE.md` already exists for every invocation path. I read the follow-up comment explaining the reasoning: `send.ts` is guarded by
  `getSession()`, so the daemon should have been started already and `start.ts` should have created `CLAUDE.md`.
     - **Impact:** that is still only an assumption, not a runtime guarantee enforced by the code. The current branch tip still leaves `send.ts` calling `runUserMessage()` without first ensuring `CLAUDE.md` exists. After this PR, if that
  invariant is ever false, the managed prompt templates are dropped for the `send` path.
     - **Fix:** add `ensureProjectClaudeMd()` to `send.ts` before `runUserMessage()`. I agree with the author’s comment that this is the better fix than restoring a second fallback path.

  2. **Missing required version bumps for shipped runtime changes**
     - **Problem:** this PR changes shipped code under `src/runner.ts` but does not update `.claude-plugin/plugin.json` or `.claude-plugin/marketplace.json`.
     - **Impact:** under the current version-guard workflow rules on `master`, this should require both version bumps.

  ## Stale-Base / Branch State

  - This branch is still stale relative to current `master`. It does not share the current `master` merge-base that the updated `#92` and `#93` branches now do.
  - Please rebase or merge current `master` before final review so the PR is evaluated against the current branch state and workflow configuration.

Directionally, I still think the core idea in this PR is right: if `ensureProjectClaudeMd()` is the canonical materialisation step, then reading both the raw templates and `CLAUDE.md` on every invocation is redundant and likely duplicates prompt content. But the current branch tip still needs the `send.ts` cleanup before that assumption is safe.

christian-drescher:
The review suggests adding ensureProjectClaudeMd() to send.ts before runUserMessage(), but this PR already implements this.

Hence, the reviewer's concern is already resolved.

(Incidentally, there is now a `CLAUDE.md` in the project root. Hence, we have even stronger guarantees now.)

TerrysPOV:
Thanks again @christian-drescher. I have combined these prompt/CLAUDE.md fixes into #145 so we can land them with the related small stability fixes while avoiding duplicate version-bump churn. I preserved attribution for your contribution in the combined branch.

PR #101, opened 2026-04-19
fix: wire sessionTimeoutMs through parseSettings and fix Discord error display

TerrysPOV:
## Summary

- **`sessionTimeoutMs` silently ignored**: `parseSettings()` didn't include `sessionTimeoutMs` in its return, so the value was stripped on every settings load. `runner.ts` used `(settings as any).sessionTimeoutMs` expecting it to pass through, but it always fell back to the hardcoded 5-minute `CLAUDE_TIMEOUT_MS`. Long-running jobs (digests, batch processing) consistently timed out at 300s regardless of what was set in `settings.json`.
- **Discord showed "Unknown error" on failures**: When Claude exits with code 1, the human-readable error is written to `stdout` as JSON, not `stderr`. The error display checked `stderr` first, which is always empty in this case, so users always saw `"Unknown error"` instead of the actual failure reason (e.g. "Autocompact is thrashing", "Prompt is too long").

## Changes

- `src/config.ts`: Add `sessionTimeoutMs: number` to `Settings` interface and `DEFAULT_SETTINGS` (30 min default); parse it in `parseSettings()` with validation
- `src/runner.ts`: Replace `(settings as any).sessionTimeoutMs || CLAUDE_TIMEOUT_MS` with `settings.sessionTimeoutMs` at both call sites
- `src/commands/discord.ts`: Fix error display order to `result.stdout || result.stderr || "Unknown error"`

## Test plan

- [ ] Set `sessionTimeoutMs: 60000` in `settings.json`, trigger a long-running job, confirm it times out at 60s not 300s
- [ ] Set `sessionTimeoutMs: 1800000`, confirm long jobs complete without premature timeout
- [ ] Trigger a Claude exit-1 condition, confirm Discord shows the actual error text instead of "Unknown error"

### DIFF
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -641,7 +641,7 @@ async function handleMessageCreate(token: string, message: DiscordMessage): Prom
     const result = await runUserMessage("discord", prefixedPrompt, threadId);
 
     if (result.exitCode !== 0) {
-      await sendMessage(config.token, channelId, `Error (exit ${result.exitCode}): ${result.stderr || result.stdout || "Unknown error"}`);
+      await sendMessage(config.token, channelId, `Error (exit ${result.exitCode}): ${result.stdout || result.stderr || "Unknown error"}`);
     } else {
       const { cleanedText, reactionEmoji } = extractReactionDirective(result.stdout || "");
       if (reactionEmoji) {
--- a/src/config.ts
+++ b/src/config.ts
@@ -61,6 +61,7 @@ const DEFAULT_SETTINGS: Settings = {
   security: { level: "moderate", allowedTools: [], disallowedTools: [] },
   web: { enabled: false, host: "127.0.0.1", port: 4632 },
   stt: { baseUrl: "", model: "" },
+  sessionTimeoutMs: 30 * 60 * 1000,
 };
 
 export interface HeartbeatExcludeWindow {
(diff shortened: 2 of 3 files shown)


christian-drescher:
I've also noticed that the hardcoded 5-minute timeout sometimes is not enough. Your PR uses two competing defaults now, though:

- You're keeping the old default `CLAUDE_TIMEOUT_MS = 5 * 60 * 1000` in `runner.ts` for `runClaudeOnce()`.

- You have added the new magic number `30 * 60 * 1000` in `config.ts` in two different locations. How about introducing a new exported constant instead (or reusing `CLAUDE_TIMEOUT_MS`)?

TerrysPOV:
Good catch, @cdrescher — you're right on both points. I've pushed a follow-up commit that:

1. Introduces a single exported `DEFAULT_SESSION_TIMEOUT_MS` constant in `config.ts` (30 min)
2. Removes the old local `CLAUDE_TIMEOUT_MS` constant from `runner.ts` entirely — it was serving the same purpose as the default parameter for `runClaudeOnce()`, so now both `DEFAULT_SETTINGS`, the `parseSettings()` fallback, and the `runClaudeOnce()` default parameter all reference the same value

No more competing magic numbers.

TerrysPOV:
Closed as included in PR #129

PR #107, opened 2026-04-24
feat: add agent-scoped job support at agents/<name>/jobs/<job>.md

TerrysPOV:
### What this does

Adds a second location for scheduled job files: `agents/<name>/jobs/<job>.md` at the project root.

Jobs placed here are automatically discovered by `loadJobs()` and run with a per-agent session at `agents/<name>/session.json`. The legacy `.claude/claudeclaw/jobs/` location continues to work unchanged.

### Why

Claude Code's CLI has a hardcoded write-protection list that blocks writes to `.claude/` paths even with `--dangerously-skip-permissions`. Discord/Telegram requests that ask Claude to create a cron job file at `.claude/claudeclaw/jobs/` fail with "system is blocking writes."

The `agents/<name>/jobs/` path is at the project root — outside `.claude/` — and is not subject to this protection.

### Changes

| File | Change |
|------|--------|
| `src/jobs.ts` | Add `agent`, `label`, `enabled` fields to `Job` interface; scan `agents/*/jobs/*.md` after legacy scan; directory location overrides frontmatter `agent:` value |
| `src/sessions.ts` | Add optional `agentName` param to all session functions; agent sessions stored at `agents/<name>/session.json` |
| `src/runner.ts` | Thread `agentName` through `execClaude()` and `run()` |
| `src/commands/start.ts` | Pass `job.agent` to `run()` at the cron fire site |
| `src/__tests__/jobs.test.ts` | 13 tests covering both scan locations, field parsing, directory authority, `enabled: false`, write-protection path property |

Net: +261 lines (+183 tests, +78 source). No new dependencies. No new processes.

### Backwards compatibility

Existing jobs in `.claude/claudeclaw/jobs/` load and run unchanged. All callers without `agentName` behave identically to before.

### End-to-end verified

Daemon loaded jobs from both locations in an isolated temp project; agent-scoped job fired at the cron boundary: `Running: testbot/frequent (new session)`.

### Test plan

- [x] `bun test src/__tests__/jobs.test.ts` — 13/13 pass
- [x] `bun x tsc --noEmit` — no new errors
- [x] Daemon end-to-end: agent job fired at cron boundary in isolation

### Relationship to the `.claude/` protection fix

If a fix moving `JOBS_DIR` out of `.claude/claudeclaw/jobs/` has already merged or is in progress, this PR is additive — it provides a semantically richer alternative path alongside that change.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,8 +8,13 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.0",
-      "keywords": ["cron", "heartbeat", "scheduler", "daemon"],
+      "version": "1.0.1",
+      "keywords": [
+        "cron",
+        "heartbeat",
+        "scheduler",
+        "daemon"
+      ],
       "category": "productivity"
     }
   ]
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/__tests__/jobs.test.ts
+++ b/src/__tests__/jobs.test.ts
@@ -1,0 +1,183 @@
+import { describe, test, expect, beforeEach, afterAll } from "bun:test";
+import { mkdir, writeFile, rm } from "fs/promises";
+import { join } from "path";
+
+const TEST_ROOT = join(import.meta.dir, "../../test-sandbox-jobs");
+const LEGACY_JOBS_DIR = join(TEST_ROOT, ".claude", "claudeclaw", "jobs");
+const AGENTS_DIR = join(TEST_ROOT, "agents");
+
+async function resetSandbox() {
+  await rm(TEST_ROOT, { recursive: true, force: true });
+  await mkdir(LEGACY_JOBS_DIR, { recursive: true });
+  await mkdir(join(AGENTS_DIR, "suzy", "jobs"), { recursive: true });
+  await mkdir(join(AGENTS_DIR, "reg", "jobs"), { recursive: true });
+}
+
+afterAll(async () => {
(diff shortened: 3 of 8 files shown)


squirblej:
Hi @TerrysPOV — I opened #113 (configurable `jobsDir` setting) for the same root problem before noticing this PR. These are two different approaches to the same issue; wanted to flag the overlap and ask how the maintainer would like to combine them.

**#107 (yours):** Convention-based `agents/<name>/jobs/` path at project root. Richer — agent-scoped sessions, `enabled` flag, tests. The path is outside `.claude/` so write-protection does not apply.

**#113 (mine):** Explicit `jobsDir` setting in `settings.json`. Simpler — single configurable path, works for any layout.

A few gaps I noticed in #107 that might be worth addressing before merge:

- `clearJobSchedule()` still writes to the legacy `JOBS_DIR` path, so non-recurring agent-scoped jobs would fail to clear their schedule after firing (silent no-op rather than updating the correct file)
- `status.ts` still reads only from the legacy `JOBS_DIR`, so agent-scoped jobs do not appear in `claudeclaw status` output

The two approaches are not mutually exclusive — `jobsDir` could serve as a configurable root for the agent scan, or they could coexist as complementary lookup paths. Happy to defer to whatever direction the maintainer prefers.

TerrysPOV:
@squirblej, Thanks for catching those — both gaps are real and now fixed in the latest push.
                                                                                                                                                                                               
clearJobSchedule fix: Added a resolveJobPath() helper that detects the agentName/label format and routes to agents/<name>/jobs/<label>.md. Legacy names are unaffected.
                                                                                                                                                                                               
status.ts fix: Status now scans both JOBS_DIR and agents/*/jobs/ so all jobs appear in claudeclaw status output.                                                                             
                                                                                                                                                                                               
On #113: Agreed the two approaches are complementary. A configurable jobsDir as a root for the agent scan makes sense. Happy to adapt this PR however @moazbuilds prefers — they can coexist (legacy + configurable root + agent-scoped), or #113 can subsume this one if that's cleaner.

TerrysPOV:
Superseded by #134, which rebases and cherry-picks the current #107 changes onto current master together with #131 and fresh version guard bumps.

PR #110, opened 2026-04-25
fix(runner): auto-recover from stale --resume session IDs

lunklin:
## Summary

When Claude Code's CLI returns `Error: No conversation found with session ID: <id>` (because the cached session was cleared, expired, compacted away, or moved to another machine via sync), claudeclaw currently keeps that dead ID in `session.json` and every subsequent message produces the same error. The only recovery today is to manually delete the session file or run a reset command — Telegram/Discord users typically just see `Error (exit 1): No conversation found with session ID: …` and have no way out.

This PR teaches `runner.ts` to detect that error in both code paths and recover automatically.

## What it does

In both `execClaude` (print mode) and `streamClaude` (stream-json mode), after the spawned `claude` process exits non-zero, scan stderr (and stdout for the print path) for the pattern `No conversation found with session ID`. When matched:

1. Backup or remove the dead session
   - Global session → `backupSession()` (preserves the dead ID under `session_<n>.backup`).
   - Thread session → `removeThreadSession(threadId)` (drops the dead entry from `sessions.json`).
2. Rebuild the argv: strip `--resume <id>` and (in the print path) force `--output-format json` so we capture the brand-new `session_id` Claude returns.
3. Retry once with the same model/security config. If it succeeds, persist the new session via `createSession()` / `createThreadSession()` and return its output as the normal reply.

Two small bookkeeping fixes guard against double-counting after recovery:
- The auto-compact-on-timeout block is skipped when `recoveredFromStale` is true (otherwise `/compact` would target the dead ID and fail the same way).
- Turn-count increment is skipped after recovery (the new session was just created with `turnCount = 0`; the next call resumes normally and counts from there).

No new dependencies, no behavior changes for the normal happy path or for rate-limit handling.

## Repro before the fix

\`\`\`bash
cat > .claude/claudeclaw/session.json <<JSON
{\"sessionId\":\"deadbeef-dead-dead-dead-deaddeaddead\",\"createdAt\":\"2026-04-25T04:08:36.853Z\",\"lastUsedAt\":\"2026-04-25T06:56:34.288Z\",\"turnCount\":5,\"compactWarned\":false}
JSON
# Then send any message via Telegram/Discord/CLI:
#   → \"Error (exit 1): No conversation found with session ID: deadbeef-...\"
# And every subsequent message gets the same error until session.json is hand-deleted.
\`\`\`

After the fix, the same message should log `Stale session deadbeef for <name>; recovering with a new session...`, move the dead ID to `session_<n>.backup`, and return a normal reply from a freshly created session — all in one turn.

## Test plan

- [x] \`bunx tsc --noEmit\` — no new type errors introduced (pre-existing errors in \`config.ts\`, \`ui/server.ts\`, \`whisper.ts\` and the Bun/DOM \`WebAssembly\` clash are unchanged).
- [ ] Manual repro: I'll add a screenshot/log once I've verified locally — patch was authored against my live install but not yet exercised end-to-end at PR-open time.
- [ ] Reviewers: please double-check that the recovery path interaction with auto-compact retry is what you want (currently gated by \`!recoveredFromStale\` so we don't loop on a dead ID).

### DIFF
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -1,12 +1,13 @@
 import { mkdir, readFile, writeFile } from "fs/promises";
 import { join } from "path";
 import { existsSync } from "fs";
-import { getSession, createSession, incrementTurn, markCompactWarned } from "./sessions";
+import { getSession, createSession, incrementTurn, markCompactWarned, backupSession } from "./sessions";
 import {
   getThreadSession,
   createThreadSession,
   incrementThreadTurn,
   markThreadCompactWarned,
+  removeThreadSession,
 } from "./sessionManager";
 import { getSettings, type ModelConfig, type SecurityConfig } from "./config";
 import { buildClockPromptPrefix } from "./timezone";
@@ -58,6 +59,37 @@ export interface RunResult {
 }
 
 const RATE_LIMIT_PATTERN = /you.ve hit your limit|out of extra usage/i;
+
+// Claude Code prints this when --resume references a session it no longer
+// has on disk (cleared, expired, compacted away, or moved to another machine).
+// When we see it, the cached session ID is dead and the only recovery is to
+// drop --resume and start fresh.
+const STALE_SESSION_PATTERN = /No conversation found with session ID/i;
+
+function isStaleSessionError(stdout: string, stderr: string): boolean {
+  return STALE_SESSION_PATTERN.test(stderr) || STALE_SESSION_PATTERN.test(stdout);
+}
+
+/** Strip --resume <id> from a claude argv list so it runs as a brand-new session. */
+function stripResume(args: string[]): string[] {
+  const out: string[] = [];
+  for (let i = 0; i < args.length; i++) {
+    if (args[i] === "--resume") {
+      i += 1; // skip the session id that follows
+      continue;
+    }
+    out.push(args[i]!);
+  }
+  return out;
+}
+
+/** Replace the value following --output-format (mutates a copy and returns it). */
+function withOutputFormat(args: string[], format: string): string[] {
+  const out = [...args];
+  const idx = out.indexOf("--output-format");
+  if (idx >= 0 && idx + 1 < out.length) out[idx + 1] = format;
+  return out;
+}
 
 // Serial queue — prevents concurrent --resume on the same session
 // Global queue for non-thread messages (backward compatible)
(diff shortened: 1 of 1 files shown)


lunklin:
Verified end-to-end on my install. Reproduced the failure with a planted stale session ID, then sent a normal message via `claudeclaw send`:

```
# .claude/claudeclaw/session.json BEFORE
{
  "sessionId": "75ee78dd-a477-4b31-a138-db89e97d1e18",   # known-bad
  "createdAt":  "2026-04-25T04:08:36.853Z",
  "lastUsedAt": "2026-04-25T07:55:16.620Z",
  "turnCount":  5,
  "compactWarned": false
}

$ bun src/index.ts send "say hi in 3 words"
[15:56:14] Stale session 75ee78dd for send; recovering with a new session...
[15:56:14] Session created: 613a6bd6-3783-4c83-be58-224b978be32a
[15:56:14] Done: send → .claude/claudeclaw/logs/send-2026-04-25T07-56-04-631Z.log

# .claude/claudeclaw/session.json AFTER
{
  "sessionId": "613a6bd6-3783-4c83-be58-224b978be32a",   # fresh
  "createdAt":  "2026-04-25T07:56:14.682Z",
  "lastUsedAt": "2026-04-25T07:56:14.682Z",
  "turnCount":  0,
  "compactWarned": false
}

# Dead ID was preserved, not lost:
$ ls .claude/claudeclaw/session*.backup
session_1.backup  session_2.backup  session_3.backup   # ← new
```

Reply text from the recovered session came through cleanly (`"hey there, Luke ⚡"`), exit code 0, single round-trip — the user never saw the underlying error.

TerrysPOV (changes requested):
I think the runner fix itself is directionally right, but there are two issues before I’d treat this as mergeable:

  1. This branch is not scoped just to the stale-session recovery change. It is stacked on top of the earlier heartbeat branch and even includes the empty retrigger commits from #50. In the actual diff that shows up as an unrelated src/commands/start.ts heartbeat change alongside the runner.ts work. I’d strongly prefer rebasing/restacking this so the PR contains only the stale-session recovery fix.
  2. I also couldn’t verify a clean typecheck on the branch as submitted. After bun install, bunx tsc --noEmit still fails on the separate existing src/config.ts parseSettings arity issue. That looks unrelated to this fix, but it means this branch should be rebased onto a base that already includes that config fix before final validation.

I did not find a separate blocking correctness issue in the stale-session recovery logic itself. The main path in runner.ts looks sensible: detect the dead --resume session, back it up/remove it, retry once without --resume, and avoid compact/turn-count misuse after recovery.

  I’d suggest:

  - rebase/restack this branch so it contains only the stale-session recovery work
  - rebase onto a base that already includes the parseSettings signature fix
  - add automated coverage for:
      - stale global session recovery
      - stale thread session recovery
      - stale stream-session recovery

TerrysPOV:
Thanks for this contribution, @lunklin! Stale session recovery is a real pain point — the permanent error loop with no user-visible escape path is one of those things that's invisible on happy paths but very disruptive in practice (especially for users who sync sessions across machines).

This has been ported to a fresh branch against current master as PR #168. Closing this one in favour of the clean port.

PR #111, opened 2026-04-25
fix: surface real Claude CLI errors instead of Unknown error

TerrysPOV:
Supersedes closed PR `#58`.

## Problem

Claude CLI sometimes exits non-zero with the real human-readable error detail in JSON stdout rather than stderr. In those cases, ClaudeClaw was surfacing generic fallback text such as `Unknown error` instead of the actual failure reason.

Example:

Before:
```text
Error (exit 1): Unknown error
```

After:
```text
Error (exit 1): Not logged in · Please run /login
```

## What this PR changes

This PR adds a small shared helper, `extractErrorDetail()`, in `src/messaging.ts` and uses it in the existing error-reporting paths for:
- daemon forwarding in `src/commands/start.ts`
- Telegram message handling in `src/commands/telegram.ts`
- Discord message handling in `src/commands/discord.ts`

The helper:
- prefers trimmed `stderr` when present
- otherwise tries to parse JSON `stdout`
- extracts Claude CLI error detail from:
  - `is_error/result`
  - `error.message`
  - string `error`
- falls back to raw `stdout` if parsing does not apply

This branch is intentionally narrower than the abandoned `#58` branch. It keeps only the shared error-detail extraction that is actually used today and drops the unused extra messaging helpers from the original PR.

## Additional cleanup included

I also folded in the small `parseSettings(raw, discordUserIds?)` signature fix in `src/config.ts` so the branch typechecks cleanly on current `master`. That is not the main purpose of this PR, but it was necessary to validate the branch properly.

## Testing

Validated with:
```bash
bun test src/messaging.test.ts
bunx tsc --noEmit
```

Added focused coverage for:
- `stderr` populated
- `stderr` empty with JSON stdout containing `is_error/result`
- non-JSON stdout fallback

## Notes

This PR recreates the useful part of closed PR `#58` on top of current `master`, with the scope narrowed to the actual runtime behaviour needed now.

### DIFF
--- a/docs/handoffs/pr58-recreation.md
+++ b/docs/handoffs/pr58-recreation.md
@@ -1,0 +1,66 @@
+# Handoff: Recreate Closed PR #58
+
+## Goal
+Recreate the useful part of closed PR `#58` on top of current `master` without carrying over the stale or unused parts of the original branch.
+
+## Branch
+`fix/surface-real-cli-errors`
+
+## Original PR
+- Closed PR: `#58`
+- Original commit: `2bc524e`
+- Problem statement: Claude CLI sometimes exits non-zero with the human-readable error message in JSON stdout rather than stderr, causing chat bridges to show `Unknown error`.
+
+## Scope Chosen
+Keep only the narrow behaviour that is still useful today:
+- add `extractErrorDetail()` in `src/messaging.ts`
+- use it in:
+  - `src/commands/start.ts`
+  - `src/commands/telegram.ts`
+  - `src/commands/discord.ts`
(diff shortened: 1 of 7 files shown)


TerrysPOV:
Closing this in favour of the follow-up hotfix now that the regression has been patched on top of current master: #122.

The earlier handling here was the safer one. The issue turned out to be that Array.isArray(discordUserIds) is too broad on its own, because an empty extracted array can suppress the parsed JSON fallback. The hotfix restores the intended behaviour by only preferring extracted raw IDs when the extracted array is non-empty.

PR #112, opened 2026-04-25
fix(discord): join new threads at THREAD_CREATE and restore delivery on reconnect

### LINKED ISSUE #108: Discord: bot doesn't receive messages in threads (missing thread join on THREAD_CREATE)
## Bug

When a user creates a new thread in a listened channel, the bot never receives `MESSAGE_CREATE` events from that thread. The bot replies to the first message in the parent channel but is then silent in the thread.

## Root cause

The `THREAD_CREATE` gateway event handler in `src/commands/discord.ts` tracks the thread in `knownThreads` but never calls `PUT /channels/{threadId}/thread-members/@me` to join it. Discord's gateway only delivers `MESSAGE_CREATE` events for threads the bot has explicitly joined.

Current code:
```typescript
case "THREAD_CREATE":
  if (data.id && data.parent_id) {
    knownThreads.set(data.id, { parentId: data.parent_id });
    debugLog(`Thread tracked: ${data.id} (parent: ${data.parent_id})`);
  }
  break;
```

## Fix

Join the thread when its parent channel is in `config.listenChannels`:

```typescript
case "THREAD_CREATE":
  if (data.id && data.parent_id) {
    knownThreads.set(data.id, { parentId: data.parent_id });
    debugLog(`Thread tracked: ${data.id} (parent: ${data.parent_id})`);
    if (config.listenChannels.includes(data.parent_id)) {
      discordApi(token, "PUT", `/channels/${data.id}/thread-members/@me`)
        .catch((err) => console.error(`[Discord] Failed to join thread ${data.id}: ${err}`));
    }
  }
  break;
```

## Steps to reproduce

1. Configure `listenChannels` with a channel ID
2. Create a new thread in that channel
3. Send a message in the thread
4. Bot does not respond

### LINKED ISSUE #109: Discord: null bytes in copy-pasted messages cause crash
## Bug

Pasting certain content into Discord (e.g. from some terminals or editors) introduces null bytes (`\0`) into `message.content`. This causes a crash when the content is later passed as a CLI argument to `Bun.spawn`, since null bytes terminate C strings at the OS level.

## Root cause

`message.content` is used for logging and prompt assembly before it reaches the spawn call. Stripping null bytes at the argument level (too late) still leaves them in log output, causing terminal corruption, and doesn't protect other uses of the raw content string.

## Fix

Strip null bytes from `message.content` at the point of extraction in `handleMessageCreate`, before any downstream use:

```typescript
// Before:
const content = message.content;

// After:
const content = message.content.replace(/\0/g, "");
```

This keeps all downstream code (logging, prompt building, CLI args) clean without requiring defensive stripping in multiple places.

## Steps to reproduce

1. Copy text from a terminal or rich editor that may embed null bytes
2. Paste into a Discord channel the bot is listening to
3. Bot crashes / restarts

squirblej:
Fixes #108
Also fixes #109

## What breaks

Three separate bugs in the Discord message path:

**Bug 1 — null bytes crash the bot (#109)**
`message.content` is used raw for logging, prompt assembly, and eventual CLI spawn. Null bytes embedded in pasted content (e.g. from terminals or rich editors) terminate C strings at the OS level, crashing `Bun.spawn`. Stripping at the spawn boundary is too late — they already contaminate logs.

**Bug 2 — new threads go silent (THREAD_CREATE, #108)**
The `THREAD_CREATE` handler added the thread to `knownThreads` but never called `PUT /channels/{id}/thread-members/@me`. The bot therefore never received `MESSAGE_CREATE` events inside newly created threads.

**Bug 3 — reconnect/resume goes deaf (#108)**
`rejoinThreads()` did a bare `PUT` to re-establish membership after a gateway reconnect. When the bot was already a REST member, Discord treated the `PUT` as idempotent and did **not** send `THREAD_MEMBERS_UPDATE`. Without that event, the new gateway session never received message delivery for that thread.

## Fix

1. Strip null bytes from `message.content` at the point of extraction, before any downstream use.

2. In `THREAD_CREATE`: issue `PUT /thread-members/@me` when the parent channel is in `listenChannels`.

3. In `rejoinThreads()`: `DELETE` then `PUT` to force a genuine rejoin and trigger `THREAD_MEMBERS_UPDATE`.

## Validation

**Null byte path:** Send a message with embedded null bytes — bot processes it without crashing; logs are clean.

**New thread path:** Create a thread under a `listenChannels` channel, send a message — bot responds without a mention.

**Reconnect path:** Restart daemon, confirm `Rejoined thread: <id>` in logs, send message in thread — bot responds.

All three paths confirmed working in production on a live Discord guild.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.2",
+      "version": "1.0.3",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.2",
+  "version": "1.0.3",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -240,6 +240,7 @@ async function rejoinThreads(token: string): Promise<void> {
   const threadSessions = await listThreadSessions();
   for (const ts of threadSessions) {
     try {
+      await discordApi(token, "DELETE", `/channels/${ts.threadId}/thread-members/@me`).catch(() => {});
       await discordApi(token, "PUT", `/channels/${ts.threadId}/thread-members/@me`);
       if (!knownThreads.has(ts.threadId)) {
         const ch = await discordApi<{ parent_id?: string }>(token, "GET", `/channels/${ts.threadId}`);
@@ -433,7 +434,7 @@ async function handleMessageCreate(token: string, message: DiscordMessage): Prom
   const channelId = message.channel_id;
   const isDM = !message.guild_id;
   const isGuild = !!message.guild_id;
-  const content = message.content;
+  const content = message.content.replace(/\0/g, "");
 
   // Recover lost thread from sessions.json (fallback for knownThreads volatility)
   if (isGuild && !knownThreads.has(channelId)) {
@@ -1008,6 +1009,11 @@ function handleDispatch(token: string, eventName: string, data: any): void {
       if (data.id && data.parent_id) {
         knownThreads.set(data.id, { parentId: data.parent_id });
         debugLog(`Thread tracked: ${data.id} (parent: ${data.parent_id})`);
+        if (getSettings().discord.listenChannels.includes(data.parent_id)) {
+          discordApi(token, "PUT", `/channels/${data.id}/thread-members/@me`).catch((err) =>
+            console.error(`[Discord] Failed to join thread ${data.id}: ${err}`),
+          );
+        }
       }
       break;
 
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -107,7 +107,6 @@ let resumeGatewayUrl: string | null = null;
 let heartbeatAcked = true;
 let running = true;
 let discordDebug = false;
-let hasGottenReady = false;
 
 // Bot identity (populated from READY)
 let botUserId: string | null = null;
@@ -953,14 +952,6 @@ function handleDispatch(token: string, eventName: string, data: any): void {
 
   switch (eventName) {
     case "READY":
-      if (hasGottenReady) {
-        // In-process reconnect with a fresh IDENTIFY — Discord's gateway won't reliably
-        // deliver thread messages after this even with DELETE+PUT rejoin. Exit cleanly
-        // so systemd provides a fresh process, which reliably restores thread delivery.
-        console.log("[Discord] READY on reconnect — exiting for clean restart");
-        process.exit(0);
-      }
-      hasGottenReady = true;
       gatewaySessionId = data.session_id;
       resumeGatewayUrl = data.resume_gateway_url;
       botUserId = data.user.id;
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -107,6 +107,7 @@ let resumeGatewayUrl: string | null = null;
 let heartbeatAcked = true;
 let running = true;
 let discordDebug = false;
+let hasGottenReady = false;
 
 // Bot identity (populated from READY)
 let botUserId: string | null = null;
@@ -952,6 +953,14 @@ function handleDispatch(token: string, eventName: string, data: any): void {
 
   switch (eventName) {
     case "READY":
+      if (hasGottenReady) {
+        // In-process reconnect with a fresh IDENTIFY — Discord's gateway won't reliably
+        // deliver thread messages after this even with DELETE+PUT rejoin. Exit cleanly
+        // so systemd provides a fresh process, which reliably restores thread delivery.
+        console.log("[Discord] READY on reconnect — exiting for clean restart");
+        process.exit(0);
+      }
+      hasGottenReady = true;
       gatewaySessionId = data.session_id;
       resumeGatewayUrl = data.resume_gateway_url;
       botUserId = data.user.id;


TerrysPOV (changes requested):
I think the actual Discord fix here is directionally right and does cover the requested changes from both #108 and #109:

  - delete-then-put rejoin on reconnect
  - explicit join on THREAD_CREATE
  - early null-byte stripping in handleMessageCreate

  I did not find a separate blocking issue in those code changes themselves.

However, this branch is currently based on a stale master and would delete unrelated current-master CI/versioning work if merged as-is, including the new plugin/marketplace version guard workflows and helper scripts. I also couldn’t verify a clean bunx tsc --noEmit on the submitted branch because it still carries the separate existing src/config.ts parseSettings signature issue.

  So my view is:

  - the Discord fix itself looks good
  - this should be rebased/restacked onto current master before merge
  - #108 and #109 should then be closed in favour of this PR once the rebased branch is merged, not before

TerrysPOV (changes requested):
I re-reviewed the full current PR tip, and I don’t see a blocking issue in the actual Discord change.

  The branch is now properly scoped to src/commands/discord.ts, and the three substantive fixes are present:

  - delete-then-put rejoin on reconnect
  - early null-byte stripping in handleMessageCreate
  - explicit join on THREAD_CREATE for listened parent channels

  So from my side, this does address the requested changes from both #108 and #109.

I wasn’t able to verify a clean bunx tsc --noEmit on the branch as submitted, but the remaining failure is still the separate existing src/config.ts parseSettings signature issue, not this Discord logic. So the actual PR change here looks merge-safe.

One things left to do. Please review: https://github.com/moazbuilds/claudeclaw/actions/runs/24931562435/job/73010539859?pr=112 and https://github.com/moazbuilds/claudeclaw/actions/runs/24931562434/job/73010541248?pr=112 and run the two scripts and push. I'll then merge.

squirblej:
Rebased onto current master (d61c9b6) — branch now sits cleanly on top of the CI/versioning commits from #114–#117.

TerrysPOV:
Are you sure because the CI checks are still failing: https://github.com/moazbuilds/claudeclaw/actions/runs/24931562435/job/73010958281?pr=112 and https://github.com/moazbuilds/claudeclaw/actions/runs/24931562434/job/73010992442?pr=112. 

All you need to do is run:
bun run bump:plugin-version

And Push the commit.

squirblej:
Sorry @TerrysPOV missed that

TerrysPOV (approved):
All good now. Will batch your two up with another couple of PRs.

squirblej:
Added a fourth fix to this branch after further production analysis.

**Bug 4 — fresh IDENTIFY reconnects go permanently deaf**

The DELETE+PUT approach in `rejoinThreads()` reliably restores thread delivery when the gateway *resumes* (RESUMED event). However, when Discord drops the session entirely and the bot must re-IDENTIFY (resulting in a fresh READY event), even DELETE+PUT does not restore MESSAGE_CREATE delivery for existing threads in the new gateway session.

Root cause: Discord's gateway routing table for message delivery is tied to the gateway session. After a fresh IDENTIFY, the session is new, and thread message delivery is only activated by the GUILD_CREATE membership data that Discord sends at the start of that session. An in-process DELETE+PUT after READY does not reproduce that initial activation.

**Fix:** Track whether we have previously received a READY event in this process (`hasGottenReady`). If READY fires a second time (indicating a reconnect bypassed RESUME), exit with code 0 and rely on the process manager (systemd) to restart. The fresh process receives a clean GUILD_CREATE with full thread membership, and thread delivery works immediately.

RESUMED reconnects are unchanged — they continue in-process via the existing DELETE+PUT path.

TerrysPOV (changes requested):
The original Discord fixes here still look good to me:

  - delete-then-put rejoin
  - early null-byte stripping
  - explicit join on THREAD_CREATE

But I think the new follow-up reconnect workaround introduces a blocking regression risk.

Right now on a second READY, src/commands/discord.ts calls process.exit(0) and assumes systemd will restart the process. That is too broad. It kills the entire ClaudeClaw process, not just the Discord gateway, so heartbeat, jobs, web UI, Telegram, and everything else go down too. ClaudeClaw also has a standalone Discord entry point and a normal daemon start path that do not guarantee an external supervisor, so in those modes a reconnect can now silently terminate the app.

It also bypasses the normal coordinated shutdown path in start.ts, so PID/statusline/web cleanup does not run.

  So from my side, this needs one more pass before merge:

  - either make the workaround restart only the Discord gateway path
  - or route it through a proper coordinated shutdown/restart mechanism
  - but avoid a raw unconditional process.exit(0) from inside the gateway handler

I still think #108 / #109 are the right issues for this PR overall, but I would not merge the current tip until this reconnect behaviour is tightened.

squirblej:
Fair enough, too crude. Let me try to get to the bottom of the disconnects versus a simple workaround.

TerrysPOV:
Ok. you're going to need to rebase again (Main is now ahead again) and re-run the version guard scripts before submitting.

squirblej:
Updated the branch: rebased onto current master (after #121/#122 merged) and reverted the `process.exit(0)` approach from the last commit.

On further investigation the exit approach is correct as a local workaround (under systemd, it's the most reliable fix) but too blunt for a general-purpose library as you noted.

The remaining open question on reconnect delivery is harder: `THREAD_MEMBER_UPDATE` doesn't fire on a voluntary bot PUT join, so we can't use it as a gateway-level ack to confirm the DELETE+PUT registered. The original DELETE+PUT fix definitely works on clean restarts — whether it's also sufficient for in-process fresh-IDENTIFY reconnects is still unconfirmed (requires observing a natural reconnect in production). Happy to land the three confirmed fixes (#108/#109 bugs) now and track the reconnect edge case separately if that makes sense.

TerrysPOV (approved):
I re-reviewed the full current PR tip, including the latest follow-up commits, and I don’t see a blocking issue now.

The earlier in-process READY reconnect exit workaround has been reverted, which was the main concern from the previous review. What remains is the narrower fix set we wanted:

  - delete-then-put rejoin on reconnect
  - early null-byte stripping in handleMessageCreate
  - explicit join on THREAD_CREATE for listened parent channels

I also verified the current tip against master, and bunx tsc --noEmit passes on the branch now. So from my side, this looks merge-safe, and I agree this PR should close #108 and #109.

TerrysPOV:
Merged as part of PR #128

PR #113, opened 2026-04-25
feat(jobs): make jobs directory configurable outside .claude/

squirblej:
## Problem

Claude Code hard-codes `.claude/` as a sensitive directory. The permissions whitelist does not override this — `"allow": ["Write(.claude/claudeclaw/jobs/**)"]` is silently ignored.

This means `CronCreate` is blocked from creating job files when a Claude Code session is managing jobs on behalf of a user (e.g. a Discord or Telegram bot that lets users schedule tasks). The jobs directory is unreachable through the standard permission grant mechanism.

## Solution

Add a `jobsDir` setting that relocates the jobs directory to any path outside `.claude/`. When unset, the default (`<cwd>/.claude/claudeclaw/jobs/`) is preserved for full backward compatibility.

```json
{
  "jobsDir": "/home/jack/myproject/jobs"
}
```

Absolute paths are used as-is. Relative paths are resolved from `process.cwd()`.

## Changes

- `config.ts`: export `getJobsDir()` helper; add `jobsDir?: string` to `Settings`; parse `jobsDir` in `parseSettings`; use `getJobsDir()` in `initConfig()`
- `jobs.ts`: replace hard-coded path constant with `getJobsDir()` (called at runtime, not module-load)
- `ui/constants.ts`: same
- `commands/status.ts`: same

## Backward compatibility

No existing behaviour changes when `jobsDir` is absent from settings.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.3",
+      "version": "1.0.4",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/status.ts
+++ b/src/commands/status.ts
@@ -9,6 +9,7 @@ const PID_FILE = join(HEARTBEAT_DIR, "daemon.pid");
 const STATE_FILE = join(HEARTBEAT_DIR, "state.json");
 const SETTINGS_FILE = join(HEARTBEAT_DIR, "settings.json");
 
+
 function formatCountdown(ms: number): string {
   if (ms <= 0) return "now!";
   const s = Math.floor(ms / 1000);
--- a/src/ui/constants.ts
+++ b/src/ui/constants.ts
@@ -1,5 +1,4 @@
 import { join } from "path";
-import { getJobsDir } from "../config";
 
 export const HEARTBEAT_DIR = join(process.cwd(), ".claude", "claudeclaw");
 export const LOGS_DIR = join(HEARTBEAT_DIR, "logs");
--- a/src/ui/constants.ts
+++ b/src/ui/constants.ts
@@ -1,7 +1,9 @@
 import { join } from "path";
+import { getJobsDir } from "../config";
 
 export const HEARTBEAT_DIR = join(process.cwd(), ".claude", "claudeclaw");
 export const LOGS_DIR = join(HEARTBEAT_DIR, "logs");
+
 export const SETTINGS_FILE = join(HEARTBEAT_DIR, "settings.json");
 export const SESSION_FILE = join(HEARTBEAT_DIR, "session.json");
 export const STATE_FILE = join(HEARTBEAT_DIR, "state.json");


squirblej:
Note: #107 by @TerrysPOV addresses the same `.claude/` write-protection problem with a different approach — agent-scoped `agents/<name>/jobs/` paths at the project root.

Left a comment there flagging the overlap and asking how the maintainer would like to combine the two. Keeping this open in the meantime as the approaches are complementary rather than conflicting.

TerrysPOV (changes requested):
Hi @squirblej,I think the jobsDir idea itself is useful, but I found two blocking issues before merge:

  1. This branch is based on a stale master and would currently delete unrelated current-master CI/versioning work, including the new plugin/marketplace version guard workflows and helper scripts. So this needs to be rebased/restacked before it is really reviewable as a merge candidate.
  2. The jobsDir implementation is not fully wired through at runtime. getJobsDir() depends on cached settings, but status.ts and ui/constants.ts both capture const JOBS_DIR = getJobsDir() at module load time, before settings are loaded. The web UI job service then uses that frozen constant for create/delete. In practice that means these paths can keep using the default .claude/claudeclaw/jobs directory even when jobsDir is configured elsewhere.

I also couldn’t verify a clean bunx tsc --noEmit on the submitted branch because it still carries the separate existing src/config.ts parseSettings signature issue, so this should be rebased onto a base that already includes that fix before final validation.

  I’d suggest:

  - rebase/restack this branch onto current master
  - remove the top-level cached JOBS_DIR constants from the status/UI paths and resolve getJobsDir() at the point of use instead
  - rerun the standard checks on the rebased branch once the existing parseSettings fix is included
  
Once this is done I'm going to merge this, as it deals with a specific issue that my PR #107 is meant to resolve, but this needs to wait for @moazbuilds to review as it is part of a broader discussion. As you note these changes are complementary but @moazbuilds may prefer to keep this simpler for now.

squirblej:
Both issues addressed:

1. Rebased onto current master (aba3bcb).
2. Removed the frozen top-level `JOBS_DIR` constants from `status.ts` and `ui/constants.ts`. `ui/services/jobs.ts` was also importing the frozen constant from `ui/constants.ts` and has been updated. All three now call `getJobsDir()` at the point of use.

One pre-existing limitation worth flagging: `claudeclaw status` (standalone CLI) reads settings directly from file rather than going through `loadSettings()`, so `cached` is never populated in that path — `getJobsDir()` would still return the default there. This is separate from the module-load issue and unrelated to this PR. Happy to address it in a follow-up if useful, or leave it as a known gap with a note in the README.

TerrysPOV (approved):
I re-reviewed the full current PR tip, and the earlier getJobsDir() runtime bug looks addressed now.

The important fix is that the status/UI/job paths no longer cache the jobs directory at module load time. They now resolve getJobsDir() at the point of use, which means a configured jobsDir can actually take effect after settings are loaded. The branch is also now properly scoped to the jobs-dir change rather than carrying unrelated stale-base churn.

I wasn’t able to verify a clean bunx tsc --noEmit on the branch as submitted, but the remaining failure is still the separate existing src/config.ts parseSettings signature issue, not this jobs-dir logic. So from my side, the actual jobsDir change now looks good.

If you're happy to pickup the related issue in a new PR that will be great and appreciated.

TerrysPOV:
Merged as part of #121

TerrysPOV:
@squirblej I've repoened this for you. Please rebase this to the current Master and run the two guard checks and push the commit and i will merge it. apologies again for missing this.

squirblej:
Rebased onto current master and ran both guard checks:

- `bunx tsc --noEmit`: two pre-existing errors in `src/ui/server.ts` and `src/whisper.ts` — neither file is touched by this PR (confirmed via `git diff master..HEAD --name-only`)
- `bun scripts/bump-plugin-version.ts patch`: bumped to 1.0.4

Ready to merge.

TerrysPOV:
Just double checked and this was merged as part of the earlier batch merge (75257d1). Closing in favour of the already-merged batch integration.

Issue #81, opened 2026-04-07
How to update ClaudeClaw?


miraculix95:
Tried `claude plugin update claudeclaw@claudeclaw` but it says "already at the latest version (1.0.0)" — even though there have been new commits on master since I installed.

Not sure if I'm using the wrong command or if the update mechanism doesn't work with the current setup. Is there a recommended way to update to the latest version?

vinayh:
This is likely because the version wasn't bumped in the [plugin manifest](https://github.com/moazbuilds/claudeclaw/blob/master/.claude-plugin/plugin.json), so Claude Code doesn't detect any changes. I expect that uninstalling and reinstalling the plugin, and perhaps clearing the cached plugin files in between, may fix this.

EDIT: I stumbled upon [this documentation](https://code.claude.com/docs/en/plugin-marketplaces) which appears to confirm this behaviour:
> The plugin’s plugin.json must declare a different version at each pinned ref or commit. If two refs or commits have the same manifest version, Claude Code treats them as identical and skips the update.

christian-drescher:
@moazbuilds should bump versions with every meaningful change

TerrysPOV:
This is a valid issue.

The follow-up comments are right that Claude Code relies on version changes in the plugin metadata to detect updates. If shipped plugin content changes but the declared version does not, `claude plugin update` can treat newer commits as identical and skip the update.

That applies not just to `.claude-plugin/plugin.json`, but to `.claude-plugin/marketplace.json` as well.

I’ve opened two PRs to address this from both sides:
- #114 — CI guard + helper for `.claude-plugin/plugin.json`
- #115 — CI guard + helper for `.claude-plugin/marketplace.json`

The idea in both is the same:
- enforce version bumps in CI when shipped plugin files change
- provide a small helper script so maintainers do not need to edit the JSON manually
- exempt docs-only and other non-shipped changes

Once those land, this should be much harder to regress.

TerrysPOV:
PR reviewed and merged.

PR #114, opened 2026-04-25
chore(ci): enforce plugin version bumps

TerrysPOV:
## Summary

Addresses part of #81 by adding lightweight automation around ClaudeClaw plugin versioning so plugin updates are actually detectable by Claude Code.

## What this PR does

- adds a CI guard workflow that fails a PR when shipped plugin files change without a corresponding version bump in `.claude-plugin/plugin.json`
- adds a small release helper script that bumps the plugin version, defaulting to a patch increment
- exempts docs-only and other non-shipped changes from the version-bump requirement

## Guard behaviour

The workflow requires a version bump when a PR changes shipped plugin files under:
- `src/**`
- `commands/**`
- `prompts/**`
- `.claude-plugin/**`

If a PR only changes docs or other non-shipped repo files, the guard exits cleanly and does not require a plugin version update.

## Helper script

```bash
bun run bump:plugin-version
```

Defaults to a patch bump. Also supports:

```bash
bun run bump:plugin-version minor
bun run bump:plugin-version major
```

## Why

Claude Code uses `.claude-plugin/plugin.json` version changes to detect plugin updates. If shipped plugin content changes but the manifest version stays the same, `claude plugin update` may treat newer commits as identical and skip the update.

This PR makes that easier to maintain correctly and harder to forget.

## Validation

- ran the bump helper locally and confirmed it updates `.claude-plugin/plugin.json`
- sanity-checked the guard pattern against docs-only and shipped-file examples

### DIFF
--- a/.github/workflows/plugin-version-guard.yml
+++ b/.github/workflows/plugin-version-guard.yml
@@ -1,0 +1,58 @@
+name: Plugin Version Guard
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  plugin-version-guard:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Require plugin version bump for shipped plugin changes
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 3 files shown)


TerrysPOV (commented):
No blocking issue in this change. The split here makes sense: a lightweight CI guard to prevent silent missed version bumps, plus a small helper so maintainers do not need to edit .claude-plugin/plugin.json by hand every time.

I spot-checked the helper locally and confirmed bun run bump:plugin-version increments the plugin manifest version as expected. The guard scope looks right as well: shipped plugin changes under src/, commands/, prompts/, and .claude-plugin/ require a version bump, while docs-only and other non-shipped changes are exempt.

PR #115, opened 2026-04-25
chore(ci): enforce marketplace version bumps

TerrysPOV:
## Summary

Addresses the corresponding marketplace side of #81 by adding lightweight automation around `.claude-plugin/marketplace.json` versioning so marketplace metadata changes are not forgotten when shipped plugin content changes.

## What this PR does

- adds a CI guard workflow that fails a PR when shipped plugin files change without a corresponding version bump in `.claude-plugin/marketplace.json`
- adds a small release helper script that bumps the marketplace plugin version, defaulting to a patch increment
- exempts docs-only and other non-shipped changes from the marketplace version-bump requirement

## Guard behaviour

The workflow requires a marketplace version bump when a PR changes shipped plugin files under:
- `src/**`
- `commands/**`
- `prompts/**`
- `.claude-plugin/**`

If a PR only changes docs or other non-shipped repo files, the guard exits cleanly and does not require a marketplace version update.

## Helper script

```bash
```

Defaults to a patch bump. Also supports:

```bash
bun run bump:marketplace-version minor
bun run bump:marketplace-version major
```

## Why

If marketplace consumers rely on `.claude-plugin/marketplace.json` metadata, that version needs the same discipline as `.claude-plugin/plugin.json`: shipped plugin content changes should be accompanied by a corresponding marketplace metadata version change, and maintainers should not need to edit the JSON manually each time.

## Validation

- ran the marketplace bump helper locally and confirmed it updates the `claudeclaw` plugin version inside `.claude-plugin/marketplace.json`
- sanity-checked the guard pattern against docs-only and shipped-file examples

### DIFF
--- a/.github/workflows/marketplace-version-guard.yml
+++ b/.github/workflows/marketplace-version-guard.yml
@@ -1,0 +1,63 @@
+name: Marketplace Version Guard
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  marketplace-version-guard:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Require marketplace version bump for shipped plugin changes
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
+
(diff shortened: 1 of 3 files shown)


TerrysPOV (commented):
No blocking issue in the rebased version.

The rebase looks clean, and the only overlap with the earlier plugin-version PR was in package.json, which is now resolved correctly by keeping both helper scripts. The marketplace-specific version guard and bump helper are both still intact after the rebase.

I also spot-checked the helper locally and confirmed bun run bump:marketplace-version updates the claudeclaw version inside .claude-plugin/marketplace.json as expected. The guard scope still looks right as well: shipped plugin changes require a marketplace version bump, while docs-only and other non-shipped changes are exempt.

PR #117, opened 2026-04-25
docs(ci): improve version bump guidance

TerrysPOV:
## Summary

Improves the contributor guidance around the new plugin and marketplace version guards.

## What this PR does

- makes both version-guard workflows more explicit about which helper command(s) contributors should run when a check fails
- adds a pull request template section covering plugin/marketplace version metadata expectations
- adds a short README contributor note with the same guidance

## Why

The guards are working, but their failure mode should be easier to act on immediately. This change makes it clearer that:
- the helper scripts are the intended fix path
- some PRs may need one bump
- some PRs may need both
- docs-only and other non-shipped changes remain exempt

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,25 @@
+## Summary
+
+Describe the change and why it is needed.
+
+## Validation
+
+- [ ] I ran the relevant checks locally
+- [ ] I updated any docs or setup guidance affected by this change
+
+## Plugin Versioning
+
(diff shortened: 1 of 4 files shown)


TerrysPOV (commented):
I don’t see a blocking issue in this change.

This is a useful follow-up to the version-guard work: the failure messages are now much more actionable, and the contributor guidance is visible both in the PR flow and in the README. That should reduce the “why did CI fail?” churn quite a bit, especially for PRs that need both metadata bumps.

The scope also looks right. This doesn’t change the underlying enforcement policy, it just makes the expected remediation path much clearer.

PR #118, opened 2026-04-25
feat(jobs): per-job model override via frontmatter

squirblej:
## What this does

Adds a `model` frontmatter field to job files. When set, it overrides the global model (and agentic routing) for that specific job.

```markdown
---
schedule: "0 * * * *"
model: haiku
---
Check disk usage and alert if above 90%.
```

```markdown
---
schedule: "0 9 * * 1"
model: opus
---
Write a detailed weekly strategy review based on last week's logs.
```

## Why

Scheduled jobs vary widely in complexity. Routing all of them through the same model wastes money on simple tasks (monitoring, digests, quick summaries) that haiku handles fine, and may underserve complex tasks that benefit from opus. A per-job override gives fine-grained control without separate daemon instances or settings files.

## Behaviour

- `model` present in frontmatter → uses that model, **skips** agentic routing
- `model` absent → existing behaviour unchanged (global setting + agentic routing if enabled)

## Changes

| File | Change |
|------|--------|
| `src/jobs.ts` | Add `model?: string` to `Job` interface; parse `model:` from frontmatter |
| `src/runner.ts` | Add `modelOverride?` param to `execClaude()` and `run()`; apply before agentic routing check |
| `src/commands/start.ts` | Pass `job.model` to `run()` at the cron fire site |

## Validation

Added `model: haiku` to a production job file and confirmed:
- Daemon loads the job correctly (`Jobs loaded: N` in startup log, job listed with correct schedule)
- At fire time: `Job model override: haiku` appears in log before the claude invocation
- `--model haiku` is passed to the claude CLI subprocess

Backward compatible: existing job files without `model:` behave identically.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,8 +8,13 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.0",
-      "keywords": ["cron", "heartbeat", "scheduler", "daemon"],
+      "version": "1.0.1",
+      "keywords": [
+        "cron",
+        "heartbeat",
+        "scheduler",
+        "daemon"
+      ],
       "category": "productivity"
     }
   ]
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.2",
+      "version": "1.0.3",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.2",
+  "version": "1.0.3",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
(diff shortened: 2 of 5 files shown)


TerrysPOV (changes requested):
The actual per-job model override idea looks reasonable to me:

  - model is parsed from job frontmatter
  - the scheduled job path passes it through
  - the runner uses it as an override for that invocation

The problem is that this branch is based on an older master, so the current diff is not scoped just to that feature. As submitted, it would also regress already-landed config/jobs-dir fixes and the newer parseSettings state in src/config.ts, and bunx tsc --noEmit still fails on the old config signature mismatch.

  So from my side:

  - the feature itself looks directionally fine
  - but this branch needs to be rebased/restacked onto current master before it’s mergeable
  - after rebasing, I’d want the PR narrowed to just the per-job model override plus the required metadata bump

squirblej:
Rebased onto current master (which now includes the jobs-dir and parseSettings fixes from #121/#122). PR diff is now scoped to just the per-job model feature: `model?` in the Job interface, frontmatter parsing, and the runner override path.

TerrysPOV (approved):
I re-reviewed the full current PR tip, and I don’t see a blocking issue here now.

The branch is now properly rebased onto current master, and the remaining diff is the intended per-job model override plus the expected plugin/marketplace metadata bump. The implementation looks consistent end-to-end: model is parsed from job frontmatter, passed through the cron execution path, and applied in the runner before agentic routing.

I also verified the current tip against master, and bunx tsc --noEmit passes on the branch now. So from my side, this looks merge-safe.

TerrysPOV:
Merged as part of PR #128

PR #119, opened 2026-04-25
fix(config): accept extracted Discord user ids

TerrysPOV:
## Summary

Fix the `parseSettings()` signature so it accepts the extracted raw Discord user IDs from `settings.json`.

This keeps `loadSettings()` and `reloadSettings()` aligned with the actual function signature and preserves Discord snowflake IDs as raw strings from the original file text instead of relying only on parsed JSON values.

## What changed

- update `parseSettings()` to accept an optional `discordUserIds?: string[]`
- use `discordUserIds` for `discord.allowedUserIds` when provided
- fall back to the existing parsed JSON path when extracted IDs are not provided

## Why

Discord snowflake IDs can exceed JavaScript's safe integer precision. Threading the extracted raw IDs through the parser preserves them as strings and fixes the existing `parseSettings(raw, extractDiscordUserIds(rawText))` arity mismatch.

## Validation

- `bunx tsc --noEmit`

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -217,7 +217,10 @@ function parseAgenticConfig(raw: any): AgenticConfig {
   };
 }
 
-function parseSettings(raw: Record<string, any>): Settings {
+function parseSettings(
+  raw: Record<string, any>,
+  discordUserIds?: string[],
+): Settings {
   const rawLevel = raw.security?.level;
   const level: SecurityLevel =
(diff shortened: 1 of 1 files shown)


TerrysPOV:
Ovelrlaps #111

TerrysPOV:
Merged as part of #121

TerrysPOV:
Merged as part #121

PR #120, opened 2026-04-25
fix(status): load settings before resolving runtime config helpers

squirblej:
Follow-up to #113.

## Problem

`claudeclaw status` (standalone CLI) never called `loadSettings()`, so `getJobsDir()` and any other helper that depends on the settings cache always returned compile-time defaults, ignoring values configured in `settings.json`.

This was flagged as a pre-existing limitation during review of #113 — out of scope for that PR but worth addressing separately.

## Fix

One `try/catch`-wrapped `loadSettings()` call at the entry point of `status()`, before the command dispatches to `showAll()` or `showStatus()`.

The try/catch means a missing or malformed `settings.json` degrades gracefully — `status` still works, helpers just fall back to their defaults, same as today.

## Why this is a general fix, not a jobsDir-specific one

The comment in the code explains the intent: any future setting resolved via `getSettings()` or a helper like `getJobsDir()` will now work correctly in the status command without any further changes. This is architectural future-proofing, not a patch for one specific field.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.2",
+      "version": "1.0.3",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.2",
+  "version": "1.0.3",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/status.ts
+++ b/src/commands/status.ts
@@ -1,7 +1,7 @@
 import { join } from "path";
 import { readdir, readFile } from "fs/promises";
 import { homedir } from "os";
-import { getJobsDir } from "../config";
+import { getJobsDir, loadSettings } from "../config";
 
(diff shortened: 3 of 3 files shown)


TerrysPOV (changes requested):
The actual status() fix here still looks right to me. Loading settings at the start of status() is the correct way to make runtime-resolved helpers like getJobsDir() see configured values rather than compile-time defaults, and the try/catch preserves the current graceful fallback behaviour.

The main issue is that this branch is still stacked on the older #113 base rather than current master, so the PR now includes already-landed jobsDir changes as well as the old pre-#119/#122 parseSettings state. bunx tsc --noEmit still fails on that stale config signature mismatch.

  So from my side:

  - the actual status() fix looks good
  - but this should now be restacked onto current master so the PR contains just the small follow-up plus the required metadata bump

squirblej:
Restacked onto current master (cherry-picked just the `loadSettings()` fix). PR now contains only the two commits: the status fix and the required metadata bump.

TerrysPOV (approved):
I re-reviewed the full current PR tip, and I don’t see a blocking issue here now.

The branch is now properly restacked on current master, and the remaining diff is the intended status() follow-up plus the expected plugin/marketplace metadata bump. Loading settings at the start of status() is the right fix here: it makes runtime-resolved helpers like getJobsDir() use configured values instead of compile-time defaults, while the try/catch keeps the existing graceful fallback behaviour for missing or malformed settings.

I also verified the current tip against master, and bunx tsc --noEmit passes on the branch now. So from my side, this looks merge-safe.

TerrysPOV:
Merged as part of PR #128

PR #121, opened 2026-04-25
batch: merge config, statusline, env token, and jobs-dir fixes

TerrysPOV:
## Summary

Batch integration branch for the current clean versions of:
- #119 — parseSettings discordUserIds fix
- #91 — env var token precedence fix
- #86 — statusline width/rendering fix
- #113 — configurable jobs directory fix

This branch intentionally applies the functional changes together on top of current `master` and regenerates the plugin metadata bump once, rather than carrying each PR's separate bump commit.

## Included changes

- accept `discordUserIds?: string[]` in `parseSettings()` and preserve raw Discord snowflake IDs
- trim `DISCORD_TOKEN` / `TELEGRAM_TOKEN` env vars before precedence is applied
- dynamically size the statusline box to the widest rendered row
- support `jobsDir` outside `.claude/` and resolve `getJobsDir()` at point of use in job/status/UI paths

## Validation

- `bunx tsc --noEmit`

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,8 +8,13 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.0",
-      "keywords": ["cron", "heartbeat", "scheduler", "daemon"],
+      "version": "1.0.1",
+      "keywords": [
+        "cron",
+        "heartbeat",
+        "scheduler",
+        "daemon"
+      ],
       "category": "productivity"
     }
   ]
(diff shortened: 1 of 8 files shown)


vinayh:
Thanks for handling these! Suggestion for future batch PRs (or even amending the existing one): GitHub supports Co-authored-by: Name <email> trailers in commit messages, which credit each contributor on the GitHub contribution graph and commit page without adding extra commits (although `git blame` still only points to the primary author in that case). Alternatively, GitHub's per-PR "Squash and merge" produces one clean commit per PR with the original contributor as the author, so `git blame` and `git log` correctly attribute each line, at the cost of bundling.

PR #122, opened 2026-04-25
fix(config): fall back when extracted Discord ids are empty

TerrysPOV:
## Summary

Hotfix for the `parseSettings()` Discord ID guard merged via the recent batch branch.

## Problem

`allowedUserIds: Array.isArray(discordUserIds) ? discordUserIds : ...` is too broad because `Array.isArray([])` is still true. If `extractDiscordUserIds()` fails to find IDs in the raw text, the empty extracted array suppresses the JSON-parsed fallback and silently produces `[]`.

## Fix

Only prefer the extracted raw IDs when the extracted array is non-empty:

```ts
Array.isArray(discordUserIds) && discordUserIds.length > 0
```

Otherwise, fall back to the parsed JSON values.

## Why

This preserves the intended snowflake-safety behaviour without turning an extraction miss into an empty allowlist.

## Validation

- `bunx tsc --noEmit`

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.1",
+      "version": "1.0.2",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.1",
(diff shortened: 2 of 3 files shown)


TerrysPOV (commented):
This looks like the right hotfix for the regression issue I spotted. 

The earlier guard in parseSettings() was too broad: Array.isArray(discordUserIds) also returns true for [], so an extraction miss could suppress the JSON-parsed fallback and silently produce an empty allowlist. Tightening that to require a non-empty extracted array fixes the regression while preserving the intended raw-snowflake path when extraction succeeds.

I also confirmed the required plugin/marketplace metadata bumps were added for the current version-guard workflow, and bunx tsc --noEmit passes on the updated branch. From my side, this looks merge-safe.

PR #123, opened 2026-04-25
feat(discord): inline .txt and .md file attachments into the prompt

BCDel89:
## What

Discord auto-converts large text pastes (> ~2000 characters) into `.txt` file attachments. Without this change those attachments were silently ignored — the bot would see an empty message and drop it. This PR detects and inlines them into the prompt as context.

## Changes

- `src/commands/discord.ts`
  - `isTextAttachment(a)` — detects attachments with `content_type: text/*` or extension `.txt` / `.md`
  - Early-return guard updated: `!hasText` added so text-only messages aren't dropped
  - `mediaParts` label updated to include `"text"` in the console log suffix
  - Fetch block downloads the first text attachment (capped at **50 KB** with a `...[truncated]` marker to avoid overflowing context)
  - Prompt part appends `Attached text file (<filename>):\n<content>` so the model sees the file inline

## Behaviour

- Only the first text attachment is inlined (consistent with how images and voice are handled).
- Network failures are caught and logged; the model is told "downloading failed, ask to resend" rather than silently missing the file.
- Files larger than 50 KB are truncated with a visible marker so the model knows content was cut.
- `.md` files are included — useful for sharing config/code snippets alongside a question.

## Testing locally

1. Update plugin: `bun run claudeclaw-plugin.json` content changes (no version bump needed)
2. Restart claudeclaw daemon
3. In Discord, paste any block of text > 2000 chars at the bot — Discord auto-attaches it as `message.txt`
4. Bot should respond with awareness of the file's contents

## PR Plugin updates

Does only and other non-shipped changes do this PR.

### DIFF
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -340,6 +340,12 @@ function isVoiceAttachment(a: DiscordAttachment): boolean {
   return Boolean(a.content_type?.startsWith("audio/"));
 }
 
+function isTextAttachment(a: DiscordAttachment): boolean {
+  if (a.content_type?.startsWith("text/")) return true;
+  const ext = extname(a.filename).toLowerCase();
+  return ext === ".txt" || ext === ".md";
+}
+
 async function downloadDiscordAttachment(
   attachment: DiscordAttachment,
   type: "image" | "voice",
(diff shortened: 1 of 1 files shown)


TerrysPOV:
Merged as part of PR #128

PR #124, opened 2026-04-25
feat(telegram): add listenChats — respond in group chats without @mention

BCDel89:
## What

Adds a `listenChats` array to the Telegram config. Any chat ID listed there is treated as a
"listening" chat: every message triggers a response, no @mention required.

This is distinct from `allowedUserIds`, which gates *who* can interact — `listenChats` gates
*which chats* are always-on.

## Config

```json
{
  "telegram": {
    "token": "...",
    "allowedUserIds": [123456789],
    "listenChats": [-100123456789]
  }
}
```

Negative IDs are group/supergroup chats. Obtain a chat ID by forwarding a message to `@userinfobot` or checking the URL in Telegram Web.

## Changes

- `src/config.ts` — `listenChats: number[]` added to `TelegramConfig` interface and `DEFAULT_SETTINGS`; parsed via `parseSettings` with safe `Array.map(Number)` coercion
- `src/commands/telegram.ts` — `groupTriggerReason` returns `"listen_chat"` for any message in a listed chat, before the existing `return null` fallthrough

## Behaviour

- Messages from users *not* in `allowedUserIds` in a `listenChats` chat are still gated — the user-ID check runs before `groupTriggerReason`.
- An empty `listenChats: []` (the default) is a no-op; existing behaviour is unchanged.

### DIFF
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -419,6 +419,9 @@ function groupTriggerReason(message: TelegramMessage): string | null {
     }
   }
 
+  const { telegram } = getSettings();
+  if (telegram.listenChats?.includes(message.chat.id)) return "listen_chat";
+
   return null;
 }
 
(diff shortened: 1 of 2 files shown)


TerrysPOV:
Merged as part of PR #128

PR #126, opened 2026-04-25
feat(discord): upload local image paths as attachments

BCDel89:
## Summary
Detects absolute image file paths (`.png`, `.jpg`, `.jpeg`, `.gif`, `.webp`) in Claude's Discord replies and uploads them as native Discord attachments instead of leaking the raw filesystem path into chat.

## Why
When Claude generates or references a local image (e.g. via the ComfyUI skill writing to `/tmp/output.png`), the path was being posted as plain text — useless to the human on the other end. Now the file is uploaded inline.

## Changes
- New `extractImagePaths()` strips matching paths from the message body and verifies each exists on disk
- New `sendMessageWithImages()` posts via `multipart/form-data` (`payload_json` + `files[i]` parts), with 429 retry-after handling
- Reply dispatch picks the multipart path when images are found, otherwise falls through to the existing `sendMessage()`
- Uses negative-lookbehind regex so paths embedded in URLs/code aren't mis-matched

## Test plan
- [x] agent generated image (`/tmp/agent_image.png`) — posted as attachment, path stripped from text
- [x] Reply with no image path — unchanged behavior
- [x] Reply with both text and image — both arrive
- [x] Text-attachment feature (merged from `feature/discord-text-attachments`) still works

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.25",
+      "version": "1.0.26",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.24",
+      "version": "1.0.25",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.25",
+  "version": "1.0.26",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.24",
+  "version": "1.0.25",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -185,20 +185,30 @@ async function discordApi<T>(
 
 // --- Message sending ---
 
+const DISCORD_MAX_MESSAGE_LEN = 2000;
+
+function discordMessageChunks(text: string): string[] {
+  const normalized = text.replace(/\[react:[^\]\r\n]+\]/gi, "").trim();
+  if (!normalized) return [];
+  const chunks: string[] = [];
+  for (let i = 0; i < normalized.length; i += DISCORD_MAX_MESSAGE_LEN) {
+    chunks.push(normalized.slice(i, i + DISCORD_MAX_MESSAGE_LEN));
+  }
+  return chunks;
+}
+
 async function sendMessage(
   token: string,
   channelId: string,
   text: string,
   components?: unknown[],
 ): Promise<void> {
-  const normalized = text.replace(/\[react:[^\]\r\n]+\]/gi, "").trim();
-  if (!normalized) return;
-  const MAX_LEN = 2000;
-  for (let i = 0; i < normalized.length; i += MAX_LEN) {
-    const chunk = normalized.slice(i, i + MAX_LEN);
+  const chunks = discordMessageChunks(text);
+  for (let i = 0; i < chunks.length; i++) {
+    const chunk = chunks[i];
     const body: Record<string, unknown> = { content: chunk };
     // Attach components only to the last chunk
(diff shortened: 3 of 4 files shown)


TerrysPOV (changes requested):
I don’t think this is merge-safe as submitted.

This branch includes the earlier Discord text-attachment work that is already merged via #128, plus the new image-upload behaviour. The blocking issue is the image-upload logic itself.

Right now extractImagePaths() will upload any existing absolute image path that appears in Claude’s reply, not just an image generated for the current request. There is no confinement to a known safe output directory, no requirement that the file was created during this interaction, and no provenance check. In practice that creates a local-file exfiltration path for any existing image on the daemon host if the model emits its path.

  So from my side this needs changes before merge:

  - restrict uploads to a known safe generated-output location, or
  - otherwise tie uploads to files explicitly created by the current workflow/request,
  - but avoid blindly uploading any existing absolute image path from assistant text

The text-attachment part of this branch is already effectively superseded by #128. The remaining thing to review here is the image-upload behaviour, and that is the part I would block on.

TerrysPOV (changes requested):
Thanks for the update. I found one blocking issue in the current branch: it now includes the isolated guild-channel session routing changes, but slash commands still operate on the global/known-thread session rather than the same channel session used for normal guild messages. That means /reset, /compact, /status, etc. can affect or report the wrong session in a normal guild channel. Please route those command handlers through the same session key logic as message handling.

Please note there is a new master so this will need to be rebased to this.

BCDel89:
Hi @TerrysPOV — thanks for the reviews. Rebased onto latest `master` and descoped this PR to image-upload only. Two commits, +84/−5 across `src/commands/discord.ts` and `src/config.ts`.

**Addressing your reviews:**

1. **Local-file exfiltration (first review)** — fixed in `eaa0781`. `extractImagePaths()` now:
   - Resolves each path with `realpathSync` before any check (defeats symlink traversal)
   - Confines uploads to `config.imageOutputRoots` (defaults to `~/.claude/claudeclaw/outbox/discord/`) — paths outside the allowlist are left as plain text
   - Rejects files whose `mtime` predates the current request (`requestStartedAt - 30s` skew) — even within the safe dir, only files created during this turn are eligible
   - Configurable via `discord.imageOutputRoots` in `settings.json` (absolute paths only)

2. **Slash-command session routing (second review)** — moved out of this PR. That bug lives in the isolated-session code, which belongs to PR #163. Will fix it there (or in a follow-up branch dependent on it). This PR now contains zero session-routing changes.

3. **Text-attachment overlap with #128** — also gone. Only image-upload + security hardening remain.

Ready for re-review whenever you have a moment.

TerrysPOV (changes requested):
Found 1 issue:                               
                                                                                                                                                                                                                                              
  The 429 retry in sendMessageWithImages has two problems. First, retry_after isn't guaranteed to be present in the response body — if Discord omits it or sends something non-numeric, Math.ceil(undefined * 1000) produces NaN, and Bun.sleep(NaN) returns immediately, turning this into a hot spin. Second, there's no retry cap, so if Discord keeps returning 429s (e.g. a misconfigured token or sustained throttle) the function recurses forever until the stack blows.  
                                                                                                                                                                                                                                              
  Fix is straightforward — guard the delay with a fallback and add an attempt counter:                                                                                                                                                        
                                               
  async function sendMessageWithImages(                                                                                                                                                                                                       
    token: string,                                                                                                                                                                                                                            
    channelId: string,                         
    text: string,                                                                                                                                                                                                                             
    imagePaths: string[],                                                                                                                                                                                                                   
    attempt = 0,                               
  ): Promise<void> {                                                                                                                                                                                                                          
    // ...
    if (res.status === 429) {                                                                                                                                                                                                                 
      if (attempt >= 3) throw new Error(`Discord rate limit exceeded after 3 retries on ${channelId}`);                                                                                                                                     
      const data = (await res.json()) as { retry_after?: number };                                                                                                                                                                            
      const delay = typeof data.retry_after === "number" && isFinite(data.retry_after)
        ? Math.ceil(data.retry_after * 1000)                                                                                                                                                                                                  
        : 5_000;                                                                                                                                                                                                                            
      await Bun.sleep(delay);                                                                                                                                                                                                                 
      return sendMessageWithImages(token, channelId, text, imagePaths, attempt + 1);                                                                                                                                                          
    }                                          
                                                                                                                                                                                                                                              
  https://github.com/moazbuilds/claudeclaw/blob/1429af05b6170ff56928ad8d75031baaf6b7ac55/src/commands/discord.ts#L304-L311

TerrysPOV:
I have pushed a small maintainer follow-up commit to address the latest review feedback: image attachment replies now reuse the Discord 2,000-character chunking path, with the attachments sent on the final chunk so long replies do not fail the multipart message. The upload retry path only retries the attachment message, so earlier text chunks are not duplicated on a 429.

Local validation passed: `git diff --check`, `bun install`, `bun test`, and `bunx tsc --noEmit --pretty false`.

This is now merge-ready from my side and can be included and closed via the next bulk cherry-pick PR. Thanks again for the contribution.

TerrysPOV:
This has now been cherry-picked into the bulk merge PR #172, including the maintainer follow-up fixes for the image upload safety/retry/chunking paths. I am closing this source PR to avoid duplicate merge history.

Thank you @BCDel89 for the contribution.

PR #129, opened 2026-04-26
feat(jobs): per-job timeout override and session isolation

squirblej:
## Summary

- Add a `timeout` frontmatter field (in seconds) to job files, allowing long-running jobs to override the global 5-minute session timeout
- Fix job session isolation: jobs now pass `job.name` as the `threadId` so each job runs in its own persistent session rather than sharing the main conversation session

## Usage

```markdown
---
schedule: "0 19 * * *"
recurring: true
model: haiku
timeout: 900
---
Run the Tech Brief...
```

## Test plan
- [ ] Job with `timeout: 900` runs for up to 15 minutes without being killed at 5 minutes
- [ ] Job without `timeout` field still uses global `sessionTimeoutMs` / 5-minute default
- [ ] Two jobs running concurrently each get their own isolated session (verified via log `resume <id>` showing different session IDs)
- [ ] Jobs no longer resume the main conversation session

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.3",
+      "version": "1.0.4",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.4",
+      "version": "1.0.5",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -239,6 +239,8 @@ function extractReactionDirective(text: string): { cleanedText: string; reaction
 async function rejoinThreads(token: string): Promise<void> {
   const threadSessions = await listThreadSessions();
   for (const ts of threadSessions) {
+    // Skip non-snowflake keys (e.g. job names) — they are not Discord thread IDs
+    if (!/^\d{17,19}$/.test(ts.threadId)) continue;
     try {
       await discordApi(token, "DELETE", `/channels/${ts.threadId}/thread-members/@me`).catch(() => {});
       await discordApi(token, "PUT", `/channels/${ts.threadId}/thread-members/@me`);
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -670,7 +670,7 @@ async function handleMessageCreate(token: string, message: DiscordMessage): Prom
     const result = await runUserMessage("discord", prefixedPrompt, threadId);
 
     if (result.exitCode !== 0) {
-      await sendMessage(config.token, channelId, `Error (exit ${result.exitCode}): ${result.stderr || result.stdout || "Unknown error"}`);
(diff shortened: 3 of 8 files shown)


TerrysPOV (changes requested):
I think there are two issues to fix before merge.

  1. The per-job session isolation is not actually persistent in the current implementation. The scheduler now passes job.name as the threadId, but createThreadSession() explicitly skips any non-snowflake ID. Since job names are non-snowflakes, those sessions are never written to sessions.json, so later runs cannot resume them. That means this does not currently deliver the “each job has its own persistent session” behaviour described in the PR.
  2. The new timeout: frontmatter is not validated. Negative values are accepted, multiplied into milliseconds, and then passed straight into the runner timeout logic. A value like timeout: -1 can therefore cause the job to time out immediately instead of falling back or being rejected.

So from my side, I’d suggest:

  - either persist job-named sessions in sessionManager as a first-class supported case, or narrow the PR description if the intent is only in-process separation
  - validate timeout: so only sensible positive values are accepted before passing them into the runner

I did verify the current tip against master, and bunx tsc --noEmit passes on the branch.

TerrysPOV:
I've just merged 130 so you'll want to rebase to this when you take care of the changes above.

squirblej:
Both issues addressed in the latest commit:

1. **Job session persistence** — removed the snowflake guard from `createThreadSession` so job-named sessions (e.g. `podcast-curator`) are now persisted in `sessions.json` across restarts. Moved the guard into `rejoinThreads` in `discord.ts` instead, so non-snowflake keys are skipped when re-joining Discord thread membership.

2. **Timeout validation** — `timeout: -1` and similar invalid values now fall back to `undefined` (global default) rather than being passed through. Only positive finite integers are accepted.

TerrysPOV (changes requested):
I re-checked the full current PR head, including the follow-up commits pushed after the first review.

The per-job `timeout:` and session-isolation work looks good, but two fixes from `#101` are not present in the current `#129` tip:

  1. Global `sessionTimeoutMs` is still not wired through `parseSettings()`, so jobs without a per-job `timeout:` and other Claude runs still fall back to the 5-minute default.
     - `src/config.ts`: `Settings` does not include `sessionTimeoutMs`, `DEFAULT_SETTINGS` does not set it, and `parseSettings()` does not return it.
     - `src/runner.ts`: the current code still uses `(settings as any).sessionTimeoutMs ?? CLAUDE_TIMEOUT_MS`.

  2. Discord error display still prefers `stderr` over `stdout`.
     - `src/commands/discord.ts:673` still sends:
       `result.stderr || result.stdout || "Unknown error"`
     - Claude often writes the human-readable exit-1 error to `stdout`, so this still hides the useful message.

Please cherry-pick the two commits from `#101` and resolve the small overlaps in `src/config.ts` / `src/runner.ts` in favour of keeping both sets of changes:

  ```bash
  git cherry-pick -x 588e10ab5252861feefdbf4eb539e0f6dd1bbaf7
  git cherry-pick -x ef506988a5edaba13b50a2e799c9643d115fa0c7

When resolving conflicts, the intended end state is:

  - keep #129's per-job timeout override and job session isolation
  - also restore sessionTimeoutMs as a real parsed setting with a single exported default constant
  - change the Discord error display to:
    result.stdout || result.stderr || "Unknown error"

squirblej:
Cherry-picked both commits from #101 (588e10ab, ef506988) and resolved the overlaps:

- `sessionTimeoutMs` is now a properly typed field in `Settings`, with a single exported `DEFAULT_SESSION_TIMEOUT_MS` constant as the source of truth — the `(settings as any)` hack is gone
- Per-job timeout override is preserved: `timeoutMsOverride ?? settings.sessionTimeoutMs`
- `cleanSpawnEnv()` is kept (rather than the inline env spread from #101) for security
- Discord error display fixed: `stdout || stderr`
- Rebased onto current master (includes #130)

Thanks to TerrysPOV for the #101 work — both fixes were the right call.

TerrysPOV (approved):
I re-checked the full current PR tip and the previously raised issues are resolved.

What changed since the earlier review:
  - `7a21326` wires `sessionTimeoutMs` through `parseSettings()` and fixes Discord error display ordering.
  - `88d877d` consolidates the timeout default into a single exported constant.
  - The PR still preserves its intended behaviour for per-job timeout frontmatter and job session isolation.
  - The non-snowflake guard remains in place for Discord thread rejoin handling.

I did not find any remaining issues in the current diff.

TerrysPOV:
Closed as well as closing PR #101 as this has been cherry picked in and included.

squirblej:
Will this be merged @TerrysPOV ?

TerrysPOV:
Sorry, yes merged.

TerrysPOV:
Hi Pal. I'm looking for some help working through some of the PR's that were submitted a while back and are now in review with "Requested Changes" and the authors are not responding. i wondered if I assign a couple to you at a time you could see if you can implement the changes and resubmit them so we can get them merged and cleared. It's just an ask so feel free to say no.

squirblej:
Who’s that comment for @TerrysPOV ?

squirblej:
If me then yes happy to take a couple.

PR #131, opened 2026-04-28
feat(watchdog): minimal runaway-session guard with opt-in limits

TerrysPOV:
## Summary

- Adds a minimal runaway-session watchdog that stops auto-compact retries when a session has timed out too many consecutive times
- Both limits default to `null` (disabled) — **no behaviour change unless opted in**
- Configurable via `settings.json` under a new `watchdog` key
- Single new source file (`src/watchdog.ts`, 155 LOC); no new runtime dependencies

## Motivation

Without this guard, a session stuck in a timeout loop will retry indefinitely via the existing auto-compact path, draining the user's Claude Code allowance until the daemon is manually stopped.

## How it works

**Consecutive-timeout circuit breaker (primary protection):**
After `maxConsecutiveTimeouts` consecutive exit-124 timeouts on the same session, the watchdog aborts the current retry cycle and logs a warning. The counter resets on any successful exit.

**Wall-clock runtime limit (secondary, opt-in):**
`maxRuntimeSeconds` measures wall-clock time from when the session is first registered via `startSession()`. Useful as a hard cap on very long-running agentic tasks.

## Configuration (opt-in)

Add to `settings.json` to enable:

```json
"watchdog": {
  "maxConsecutiveTimeouts": 3,
  "maxRuntimeSeconds": null
}
```

Both fields default to `null` (disabled). Any absent or invalid value falls back to `null` — no silent coercion.

## Lightweight check

- No new runtime dependencies
- In-memory `Map` — entries cleared on clean exit, abort, or non-timeout failure
- 26 lines added to `runner.ts`, 8 lines to `config.ts`
- Zero changes to the happy path (watchdog block only runs when `exitCode !== 0`)

## Tests

17 tests (`bun test`) — 17 pass, 0 fail

- `parseWatchdogConfig`: valid values, rejections (negative, float, string, `Infinity`, null)
- Disabled mode never triggers
- Abort fires at exactly the configured threshold
- Counter resets on success and on non-timeout failure
- `clearSession` removes state
- `maxRuntimeSeconds` with injected clock (no real `setTimeout` in tests)
- `startSession` is idempotent (second call does not reset the clock)

## Test plan

- [x] `bun test src/__tests__/watchdog.test.ts` — 17 pass, 0 fail
- [x] `bunx tsc --noEmit` — clean
- [x] `bun run bump:plugin-version` → 1.0.5
- [x] `bun run bump:marketplace-version` → 1.0.5
- [ ] Manual: daemon starts normally with no `watchdog` key in `settings.json`
- [ ] Manual: with `maxConsecutiveTimeouts: 2`, daemon logs warning and stops after 2 consecutive timeouts

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.4",
+      "version": "1.0.5",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/__tests__/watchdog.test.ts
+++ b/src/__tests__/watchdog.test.ts
@@ -1,0 +1,180 @@
+import { describe, it, expect, beforeEach, afterEach } from "bun:test";
+import {
+  recordResult,
+  abortReason,
+  clearSession,
+  startSession,
+  parseWatchdogConfig,
+  injectClock,
+  resetClock,
+  DEFAULT_WATCHDOG_CONFIG,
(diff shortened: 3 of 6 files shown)


TerrysPOV:
Superseded by #134, which rebases and cherry-picks the current #131 changes onto current master together with #107 and fresh version guard bumps.

PR #132, opened 2026-04-28
feat(jobs): per-job retry with configurable attempts and delay

Nibbler1250:
## Summary

Adds `retry` and `retry_delay` frontmatter fields to job files. When a job exits with a non-zero code, it is automatically retried up to `retry` times before giving up until the next scheduled run.

## Usage

```markdown
---
schedule: "*/10 * * * *"
retry: 3
retry_delay: 120
---
Your job prompt here
```

- `retry`: max number of retry attempts (default: none — existing behaviour unchanged)
- `retry_delay`: seconds between retries (default: 300s / 5 min)

## How it works

- Retry state is tracked in-memory (Map) per daemon session — no external storage, no state.json changes
- On success: retry state is cleared
- On failure with retries remaining: `retryAt` is scheduled; next 60s tick checks and fires if due
- On retry exhaustion: logs a message, stops until next cron match

## Test plan

- [ ] Job with `retry: 3` that fails retries up to 3 times with logged attempts
- [ ] Job without `retry` behaves exactly as before (no regression)
- [ ] `retry_delay: 60` waits ~60s between attempts
- [ ] Successful run after 1 failure resets the counter (next failure starts fresh)
- [ ] Exhausted retries logs the exhaustion message and stops

## Notes

Retry state is intentionally in-memory only. A daemon restart clears all retry counters — this avoids stale retry debt accumulating across restarts and keeps the implementation minimal (no new files, no state.json schema changes).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.5",
+      "version": "1.0.6",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.5",
+  "version": "1.0.6",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -706,27 +706,62 @@ export async function start(args: string[] = []) {
 
   updateState();
 
+  // Track per-job retry state (in-memory, resets on daemon restart)
+  const jobRetryState = new Map<string, { failCount: number; retryAt: number }>();
+
+  function runJob(job: (typeof currentJobs)[0]) {
+    const timeoutMs = job.timeoutSeconds ? job.timeoutSeconds * 1000 : undefined;
+    resolvePrompt(job.prompt)
+      .then((prompt) => run(job.name, prompt, job.name, job.model, timeoutMs))
+      .then((r) => {
+        // Reset retry state on success
+        if (r.exitCode === 0) {
+          jobRetryState.delete(job.name);
+        } else if (job.retry && job.retry > 0) {
+          // Schedule a retry if under the limit
+          const state = jobRetryState.get(job.name) ?? { failCount: 0, retryAt: 0 };
+          state.failCount += 1;
+          if (state.failCount <= job.retry) {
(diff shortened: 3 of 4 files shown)


TerrysPOV (changes requested):
Review scope: full current `#132` head (`2e80906`) against current `master` (`b1a62d4`), including current check state. There are no PR comments to incorporate.

  ## High Priority

  1. **One-shot jobs lose their retry window after the first failure**
     - **Problem:** `runJob()` records retry state when a job fails, but the `finally` block still clears `schedule:` immediately for every non-recurring job.
     - **Impact:** after hot reload, the failed one-shot job is no longer present in `currentJobs`, because job files without `schedule:` are ignored. The retry state remains in memory, but the daemon no longer iterates that job, so the
  retry never fires.
     - **Fix:** only clear the schedule for one-shot jobs after a successful run, or after all configured retries have been exhausted.

  2. **Branch is stale and conflicts with current `master`**
     - **Problem:** current `master` now includes `#129`, and this branch is still based before that merge. A merge simulation reports conflicts in `src/commands/start.ts` and `src/jobs.ts`.
     - **Impact:** the branch must be rebased or merged carefully to preserve the `#129` behaviour: per-job timeout, job session isolation, `sessionTimeoutMs`, Discord stdout-first error handling, and the `1.0.5` plugin metadata.
     - **Fix:** rebase onto current `master` and resolve the job scheduler changes as an integration, not as a replacement.

  3. **Version guard checks are failing**
     - **Problem:** this PR changes shipped code under `src/` but does not update `.claude-plugin/plugin.json` or `.claude-plugin/marketplace.json`.
     - **Impact:** both version guard checks are correctly failing.
     - **Fix:** bump both plugin and marketplace metadata after the functional/rebase issues are resolved.

  Directionally, per-job retry is useful, but this is not ready under the repo’s stabilisation/correctness bar until the retry semantics and current-master integration are fixed.

TerrysPOV:
Closing in favour of #134.

The non-conflicting parts of this PR have been cherry-picked into #134 (commit da036a9):
- `retry` and `retryDelay` frontmatter fields on the `Job` interface
- `runJob()` helper in `start.ts` encapsulating dispatch, retry scheduling, and one-shot job cleanup

**Conflict resolved:** #134 introduces agent-scoped job dispatch (`agent:<name>` queue key and `agentName` parameter on `run()`). Rather than rebasing #132 over those changes, the retry logic has been integrated directly into #134's updated dispatch.

**Bug fixed during integration:** the original implementation did `jobRetryState.delete(job.name)` before calling `runJob()` in the retry-firing path. If the retry failed again, `runJob()` saw no existing state and started `failCount` at 0, so a permanently failing job could retry indefinitely and a one-shot job would never have its schedule cleared. Fixed by preserving the existing state entry across the retry — `runJob` reads and increments `failCount` from it naturally.

PR #133, opened 2026-04-28

Nibbler1250:
## Summary

Adds native inline keyboard button support to the Telegram integration. Claude can include a `[buttons: ...]` directive in its output to send messages with tappable buttons — no external infrastructure required.

## Usage

In any Claude response going to Telegram:

```
Deploy to production?

[buttons: ✅ Deploy | ❌ Cancel | ⏸ Later]
```

Multi-row buttons (newline separates rows, pipe separates buttons in a row):

```
[buttons: ✅ Approve | ❌ Reject
⏸ Snooze | 🔄 Retry]
```

When the user taps a button:
1. Telegram's loading spinner is acknowledged immediately  
2. The original message is edited to show `› <selected label>`
3. `[Button pressed: <label>]` is injected into the active Claude session
4. Claude responds and can send further messages (with or without new buttons)

## Implementation

- **`extractButtonsDirective`** — strips `[buttons: ...]` from Claude output, returns parsed rows
- **`sendMessageWithButtons`** — sends via `reply_markup.inline_keyboard`, falls back to plain text if HTML parse fails
- **`handleCallbackQuery`** — generic `btn:` prefix routing, message edit, session injection
- Composes with existing `[react: ...]` and `[send-file: ...]` directives — order independent
- Secretary pattern (`sec_yes/sec_no`) is untouched

## Test plan

- [ ] `[buttons: ✅ Yes | ❌ No]` renders as two side-by-side buttons in Telegram
- [ ] Multi-row: `[buttons: A | B\nC | D]` renders as 2 rows of 2 buttons
- [ ] Tapping a button edits the original message to show `› <label>`
- [ ] Claude receives `[Button pressed: <label>]` and can respond with new buttons
- [ ] Messages without `[buttons: ...]` behave exactly as before (no regression)
- [ ] Secretary pattern (`sec_yes/sec_no`) still works

## Background

We've been running a custom `pending.py` implementation of this pattern in production for 2+ months (27 trusted senders, pending-worker cron). This PR brings that capability natively into ClaudeClaw so all users benefit without needing external infrastructure. Mentioned in #14.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.4",
+      "version": "1.0.5",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/__tests__/watchdog.test.ts
+++ b/src/__tests__/watchdog.test.ts
@@ -1,0 +1,180 @@
+import { describe, it, expect, beforeEach, afterEach } from "bun:test";
+import {
+  recordResult,
+  abortReason,
+  clearSession,
+  startSession,
+  parseWatchdogConfig,
+  injectClock,
+  resetClock,
+  DEFAULT_WATCHDOG_CONFIG,
+  TIMEOUT_EXIT_CODE,
+  type WatchdogConfig,
+} from "../watchdog";
+
+const SESSION = "test-session-abc";
+const DISABLED: WatchdogConfig = { maxConsecutiveTimeouts: null, maxRuntimeSeconds: null };
+
+beforeEach(() => {
+  clearSession(SESSION);
+  resetClock();
+});
+
+afterEach(() => {
+  resetClock();
+});
+
+// --- parseWatchdogConfig ---
+
+describe("parseWatchdogConfig", () => {
+  it("returns defaults for missing input", () => {
+    expect(parseWatchdogConfig(undefined)).toEqual(DEFAULT_WATCHDOG_CONFIG);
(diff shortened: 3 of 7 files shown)


TerrysPOV (changes requested):
Review scope: full current `#133` head (`aedc120`) against current `master` (`b1a62d4`). There are no PR comments to incorporate.

  ## High Priority

  1. **Telegram button callbacks bypass the configured allowlist**
     - **Problem:** normal Telegram messages check `config.allowedUserIds`, but `handleCallbackQuery()` does not perform the same authorisation check before processing `btn:` callbacks.
     - **Impact:** any user who can interact with a bot message containing inline buttons can inject `[Button pressed: <label>]` into the Claude session, even if they would not be allowed to send normal messages.
     - **Fix:** apply the same allowlist check to callback queries using `query.from.id` before handling both generic button callbacks and secretary callbacks, unless there is a deliberate separate policy for callbacks.

  2. **Button labels are used directly as Telegram `callback_data`**
     - **Problem:** `sendMessageWithButtons()` builds `callback_data` as `btn:${label}` from arbitrary Claude output.
     - **Impact:** Telegram limits callback data to 64 bytes, so long or multi-byte labels can make message sending fail. It also makes callback routing depend on display text rather than a bounded, stable token.
     - **Fix:** store a short generated button id in `callback_data` and map it back to the label/server-side context, or enforce byte-length validation/truncation with a safe mapping.

  3. **Branch is stale and conflicts with current `master`**
     - **Problem:** current `master` includes `#129`, but this branch is based before that merge. A merge simulation reports a conflict in `src/config.ts`.
     - **Impact:** the branch must be rebased carefully to preserve the current `sessionTimeoutMs`, per-job timeout/session isolation, and related `#129` changes.
     - **Fix:** rebase onto current `master` and resolve the config/runner integration.

  4. **PR scope mixes unrelated features**
     - **Problem:** this PR is titled Telegram inline buttons, but it also includes the watchdog feature, config changes, runner integration, and watchdog tests.
     - **Impact:** this is too broad for a stabilisation-first repo and makes review/rollback harder.
     - **Fix:** split watchdog into its own PR, or remove it from this branch and keep this PR focused on Telegram inline buttons.

  ## Check State

  - `claude-review` is failing.
  - Local `bunx tsc --noEmit` passes, but this branch should not merge until the behavioural issues, stale-base conflict, and failed checks are resolved.

Nibbler1250:
Closing in favour of a clean replacement PR that addresses all review feedback: split watchdog out, fixes allowlist enforcement on callback queries, and uses stable button IDs instead of raw labels.

PR #137, opened 2026-04-29
feat(plugin): Phase 2 — per-agent runtime isolation for plugin spawns

TerrysPOV:
## Summary

Follow-up to #136. Implements per-agent plugin isolation so that plugins installed with `-s project` scope land in and are resolved from the individual agent's directory, not the shared deployment root.

- **`runner.ts`**: Named agent spawns now use `agents/<name>/` as their working directory. `ensureAgentDir(agentName)` creates the directory on first use. Claude Code resolves `.claude/plugins/` and `.claude/skills/` relative to `cwd`, so plugins installed per-agent are visible only to that agent.
- **`plugin-cli.ts`**: `runPluginCli` accepts an optional `cwd` to scope the install subprocess.
- **`plugin-wizard.ts`**: When an `agentName` is known, the scope prompt shows `agents/<name>/.claude/plugins/` for option 2. The install-confirm step passes the agent's directory as `cwd` so the install lands in the right place. Confirmation message reports the actual install location.
- **`discord.ts`**: `knownThreads` extended to store `agentName` alongside `parentId`. Populated at thread creation (hire flow) and on gateway events: `THREAD_CREATE`, `THREAD_UPDATE`, `THREAD_LIST_SYNC`. The wizard context now carries `agentName` for any thread with a known name.

Telegram has no agent-threading concept and is intentionally unchanged.

## Test plan

- [ ] `bun test` — 61 tests pass, 0 fail
- [ ] `bunx tsc --noEmit` — clean
- [ ] Manual: hire an agent thread in Discord, open `/plugin` in that thread → scope prompt shows `agents/<name>/`
- [ ] Manual: install with project scope in an agent thread → plugin lands in `agents/<name>/.claude/plugins/`
- [ ] Manual: global session DM → scope prompt shows `.claude/plugins/` (no agent scoping)
- [ ] Manual: named agent Claude spawn uses `agents/<name>/` as cwd — existing session, job, heartbeat behaviour unchanged

PR #138, opened 2026-04-30

Nibbler1250:
## Summary

- Adds `[buttons: Label A | Label B]` directive: Claude can now output inline Telegram buttons directly
- Pipe-separated labels = columns; newlines = rows
- **Full rewrite on current master** (rebased onto `af0afc5` / `1.0.12`)

## Fixes from previous review (Terry @TerrysPOV)

**1. Buttons-only directives were dropped**
`sendMessageWithButtons()` is now called when `buttonRows` is truthy, regardless of whether `cleanedText` is empty. Uses `\u200B` (zero-width space) as the message body when there is no text — satisfies Telegram's non-empty text requirement.

**2. Button IDs reused after daemon restart**
Unknown `btn:` IDs (from before a restart) are now **rejected** with a `show_alert: true` user-visible error ("Ce bouton a expiré — relance la conversation."). The raw ID is never injected as a label.

**3. Branch rebased on current master**
Fresh branch from `af0afc5` (master @ `1.0.12`). Single clean commit — no conflict markers, no stale history.

## Also included

- `callApi` HTTP timeout via `AbortSignal.timeout` (long-poll + 15s buffer)
- Allowlist enforcement on callback queries (same policy as regular messages)
- `sendMessageWithButtons` respects 4096-char chunking — buttons attach to final chunk only

## Checks

- `bunx tsc --noEmit` passes on `telegram.ts` (pre-existing error in `src/ui/server.ts` is unrelated)
- `git diff --check origin/master...HEAD` passes
- Version bumped to `1.0.13`

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.24",
+      "version": "1.0.25",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.24",
+  "version": "1.0.25",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -463,13 +463,49 @@ function extractButtonsDirective(text: string): { cleanedText: string; buttonRow
 
 // Map short button IDs to labels for callback routing (in-memory, per-process).
 // IDs are never recycled within a process lifetime — the counter is strictly monotonic.
-const buttonLabelMap = new Map<string, string>();
+// Entries carry a creation timestamp so we can evict stale ones; otherwise a long-running
+// daemon would accumulate every button label ever generated and leak memory unbounded.
+type ButtonEntry = { label: string; createdAt: number };
+const buttonLabelMap = new Map<string, ButtonEntry>();
 let _buttonCounter = 0;
+const BUTTON_TTL_MS = 24 * 60 * 60 * 1000; // 24h is well past any reasonable user dwell
+const BUTTON_MAX_ENTRIES = 5000; // hard cap as a safety net for flood scenarios
+
+function pruneExpiredButtons(now: number = Date.now()): void {
+  for (const [id, entry] of buttonLabelMap) {
+    if (now - entry.createdAt > BUTTON_TTL_MS) {
+      buttonLabelMap.delete(id);
+    }
+  }
+  // Hard cap defense: if a flood fills the map within TTL, drop oldest insertions
+  // (Map preserves insertion order) until back under the cap.
+  if (buttonLabelMap.size > BUTTON_MAX_ENTRIES) {
+    const overflow = buttonLabelMap.size - BUTTON_MAX_ENTRIES;
+    let dropped = 0;
+    for (const id of buttonLabelMap.keys()) {
+      if (dropped >= overflow) break;
+      buttonLabelMap.delete(id);
+      dropped++;
+    }
+  }
+}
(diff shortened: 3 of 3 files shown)


TerrysPOV (changes requested):
Review scope: full current PR head `c056048` against current `master` `af0afc5`, which includes PR #137. I also checked PR comments/reviews; there are no inline comments or prior reviews on this PR.

  ## Medium Priority

  1. **User-facing expiry message is not in project language**
     - **Problem:** expired button callbacks return `Ce bouton a expiré — relance la conversation.`
     - **Impact:** this is user-facing text and conflicts with the repo instruction to use UK English for user-facing copy.
     - **Fix:** use an English message, e.g. `This button has expired. Please continue the conversation.`

  ## Merge Readiness

  No blocking code findings from this pass. The previous PR #135 issues appear addressed:
  - buttons-only directives now route through `sendMessageWithButtons()`
  - unknown/expired `btn:` IDs are rejected rather than injected
  - branch is rebased on current `master`
  - plugin and marketplace versions are bumped `1.0.12 -> 1.0.13`

  Verification:
  - `bun install --frozen-lockfile` passed
  - `bun test` passed: 61 tests
  - `bunx tsc --noEmit` passed
  - `git diff --check origin/master...HEAD` passed
  - local version-guard equivalent confirms shipped files changed and both plugin/marketplace versions changed `1.0.12 -> 1.0.13`

  GitHub check status:
  - `claude-review` passed
  - `plugin-version-guard` and `marketplace-version-guard` triggered but are `action_required` with no jobs, so they have not actually passed on GitHub yet. These should be approved/rerun before merge.

Nibbler1250:
Fixed in latest commit (4a01a2f): expired button message changed to English — *"This button has expired. Please continue the conversation."*

Nibbler1250:
Rebased on current master (`dfc360c`) and bumped version to `1.0.14` (after PR #139 landed on master at 1.0.13). English expiry message fix from previous commit is included. Branch is clean.

TerrysPOV (changes requested):
Thanks for the update. The previous review finding about the expired-button message is fixed.

I found one remaining implementation issue: `buttonLabelMap` stores every generated button label and never evicts entries after use or expiry. Since claudeclaw runs as a daemon, repeated `[buttons: ...]` responses will leak memory over time. Please add TTL eviction or delete callback IDs after they are used.

Also, the branch is now conflicting with current `master`. It was last rebased on `dfc360c`, but `master` has since advanced through #145/#147. Please rebase/clean-port again and make sure the conflict resolution preserves the later Telegram/fallback/plugin work.

Local verification on the current PR head passed: `bun test`, `bunx tsc --noEmit`, and `git diff --check origin/master...HEAD`.

Nibbler1250:
Hey @TerrysPOV — pushed the fix for the buttonLabelMap memory leak you flagged.

**Changes (commit `ef4a136`):**
- Wrapped each entry as `{ label, createdAt }` and added a 24h TTL.
- New `getButtonLabel()` helper self-evicts stale entries on access.
- `makeButtonId()` runs an opportunistic O(n) prune every 100 buttons (no separate timer).
- Hard cap of 5000 entries as a flood safety net (oldest insertions drop first since Map preserves insertion order).
- Delete the entry immediately after the callback is acked, since buttons are one-shot.

Also rebased onto current master — resolved the merge conflict in `handleMessage` so the voice + buttons + sendFile + react pipelines all coexist, and bumped to `1.0.25` over master’s `1.0.24`.

`bun test` passes (95/95). Ready for another look when you have a moment.

TerrysPOV:
I have pushed a small maintainer follow-up commit to address the review feedback: button callback responses now run the voice directive path as well as text, buttons, and files. Local validation passed: git diff --check, bun install, bun test, and bunx tsc --noEmit --pretty false. This is now merge-ready from my side and will be included and closed via the next bulk cherry-pick PR. Thanks again for the contribution.

TerrysPOV:
This has now been cherry-picked into the bulk merge PR #172, including the maintainer follow-up fix for voice directives after button presses. I am closing this source PR to avoid duplicate merge history.

Thank you @Nibbler1250 for the contribution.

PR #141, opened 2026-04-30
fix(discord): null-safe optional chaining on message.mentions

squirblej:
Closes #88 — narrow version addressing TerrysPOV's review feedback.

## What changed

Single character fix: `message.mentions.some(...)` → `message.mentions?.some(...)` in `guildTriggerReason()`.

Discord gateway `MESSAGE_CREATE` events occasionally omit the `mentions` array entirely. The missing property caused an uncaught `TypeError` that put systemd-managed daemons into a restart loop.

## What was dropped from #88

- **`sessionTimeoutMs` changes** — already present on current master via the `parseSettings` work merged with #129. Carrying them forward would be a no-op at best and a conflict at worst.
- **`.gitignore` change** — the original PR had a missing newline that produced `AGENTS.md__pycache__/` as a single entry, which would corrupt `.gitignore`.

## Verification

`bunx tsc --noEmit` passes on this branch.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.13",
+      "version": "1.0.14",
       "keywords": [
         "cron",
         "heartbeat",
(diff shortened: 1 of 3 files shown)


TerrysPOV:
Thanks @squirblej for the Discord null-safety fix. I have combined the functional change into #145 so we can land it with the related small stability fixes in one clean branch. Attribution is preserved in the combined branch.

PR #142, opened 2026-04-30
fix(runner): prevent memory leaks causing OOM in long-running daemons

squirblej:
Closes #67 — addresses TerrysPOV's review feedback.

## Changes

**Queue chain reference reset** — `task.catch(() => {})` → `task.then(() => {}, () => {})` in both `globalQueue` and `threadQueues`. The catch form passes through fulfilled values, keeping a reference chain to every prior task result and preventing GC from reclaiming them.

**Timer cleanup** — `clearTimeout(timeoutId)` called in both success and error paths of `runClaudeOnce`. The original code left the timer live after the race resolved, leaking a timer reference per invocation.

**Capped stream collection with pipe drain** — replaces unbounded `new Response(proc.stdout).text()` in both execution paths (`runClaudeOnce` and the stream-json path added in #139). The `collectStream` helper caps output at 10 MB, but unlike the original PR #67 implementation, it **continues reading and discarding** after the cap rather than breaking. Breaking left the child blocked on a full pipe with no timeout left to recover — this version drains the pipe so the child exits normally.

The stream-json `readStderr` path (added after #67 was submitted) had the same unbounded read and an unprotected timeout, so both are fixed here for consistency.

## Background

In production (24/7 systemd daemon), these leaks caused the bun process to grow from ~43 MB to 23 GB over several days before OOM kill. After applying fixes, memory stays stable at ~43 MB.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.13",
+      "version": "1.0.14",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.13",
+  "version": "1.0.14",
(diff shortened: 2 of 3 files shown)


TerrysPOV:
Thanks @squirblej for tracking down and contributing the long-running daemon memory-leak fixes. I have combined the functional change into #145 and omitted the standalone version-bump commit so the release metadata can be handled once. Attribution for the original commit and your PR contribution is preserved in the combined branch.

PR #143, opened 2026-04-30
fix(runner): per-agent fallback sessions with persistent context

squirblej:
## Problem

When the primary Claude provider is rate-limited, the fallback retried using the same `args` array — which included `--resume <primary-session-id>`. The fallback provider has no knowledge of that session ID, so it either errored or silently started a fresh session while discarding context.

Additionally, error messages surfaced to users via Telegram/Discord were always drawn from stderr, but Claude CLI sometimes writes structured error details (auth failures, quota messages) into stdout in JSON format instead.

## Changes

**Per-agent fallback sessions (`sessions.ts`)**

Adds `getFallbackSession`, `createFallbackSession`, `incrementFallbackTurn`, and `resetFallbackSession`, all scoped per-agent, mirroring the existing primary session pattern. Fallback sessions persist across rate-limit events so the fallback provider accumulates context (same as primary).

**Rewritten fallback block (`runner.ts`)**

When falling back:
- Builds a fresh `fallbackArgs` array (no primary `--resume` ID)
- Loads the existing fallback session for the agent (if any) and resumes it, or starts a new one
- Persists/increments the fallback session after a successful fallback run
- Imports `getFallbackSession`, `createFallbackSession`, `incrementFallbackTurn` from sessions

**`extractErrorDetail` (`messaging.ts`)**

New helper that accepts `{ stdout, stderr }` and returns the most useful error string: prefers stderr, then tries to parse JSON stdout for structured error fields (`is_error`/`result`, `error.message`, `error`), falls back to raw stdout. Three unit tests in `src/__tests__/messaging.test.ts`.

**Error forwarding (`start.ts`, `discord.ts`, `telegram.ts`)**

Replaces `result.stderr || "Unknown"` and `result.stdout || result.stderr || "Unknown error"` with `extractErrorDetail(result) || "Unknown [error]"` at all four error-reporting sites.

## Addresses

Closes #76

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.14",
+      "version": "1.0.15",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.13",
+      "version": "1.0.14",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.14",
+  "version": "1.0.15",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.13",
+  "version": "1.0.14",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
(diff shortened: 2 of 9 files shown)


TerrysPOV (changes requested):
Thanks @squirblej. I reviewed this against the new combined direction in #145.

There is one blocker before we can cherry-pick this into the next combined PR:

  When the primary provider rate-limits on a new session and the fallback succeeds, the fallback `exec.sessionId` can still flow through the normal new-session persistence path and be saved as the primary/thread session. That means a later primary run may try to resume a session ID that belongs to the fallback provider.

  Could you please update the fallback path so that, when the successful result came from fallback, it stores only the fallback session and skips primary/thread session creation?

  One related reset issue: `resetFallbackSession()` is added but not wired into `/reset`, so fallback context can survive a user-visible fresh-session reset. Please include that as well.

After that, please rebase onto the new `master` request review again. Thanks for the contribution and the help. it is highly valued.

squirblej:
Both blockers addressed and rebased onto master (post #145):

**Fallback session ID leaking into primary** — added `&& !usedFallback` to the new-session persistence condition in `execClaude`. The fallback's session ID now stays in the fallback session only; the primary/thread session creation block is skipped entirely when the run was served by the fallback.

**`/reset` not clearing fallback** — both the Discord slash command and the Telegram `/reset` command now call `resetFallbackSession()` alongside `resetSession()`. Added `resetFallbackSession` to the imports in both files.

Ready for re-review.

TerrysPOV:
Thanks @squirblej. I have combined the fallback-session work from this PR into #147, including the follow-up fix that prevents fallback session IDs from being persisted as primary sessions and the /reset fallback cleanup. Closing this PR in favour of the combined cherry-pick PR.

PR #158, opened 2026-05-01
feat(telegram): delegate voice transcription to Claude instead of local whisper

TerrysPOV:
## Summary

- Removes the whisper dependency for voice transcription in Telegram voice messages
- Instead of transcribing locally, claudeclaw now sends the voice file path directly to Claude and asks it to transcribe
- Reduces setup complexity (no whisper binary needed) and leverages Claude's native audio understanding

## What changed

`src/commands/telegram.ts`:
- Removed `import { transcribeAudioToText } from "../whisper"`
- Removed the transcription block that called whisper
- Voice prompts now pass the file path to Claude with a transcription instruction

## Attribution

Ports #35 by @iammrwrong — adapted onto current master.

## Validation

- [x] `bunx tsc --noEmit` passes
- [x] `bun run bump:plugin-version && bun run bump:marketplace-version` run

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.20",
+      "version": "1.0.21",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.20",
+  "version": "1.0.21",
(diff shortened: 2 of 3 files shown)


TerrysPOV:
Closing this PR. The delegation approach doesn't work: the Claude CLI invocation is a text prompt and has no mechanism to attach or pass a binary audio file for transcription. Passing the file path as a string asks Claude to "use an available transcription tool" — but no such tool exists in the CLI invocation context, so this produces worse results than whisper.

A real fix would require either (a) integrating a cloud STT API (e.g. the `stt.baseUrl` path already in settings) or (b) native audio-file attachment support in the Claude CLI. Until then, the whisper path is the correct one to keep.

The original PR #35 by @iammrwrong had the right intent (remove the native binary dependency) but the replacement approach needs more infrastructure. Closing in favour of continued use of the whisper path.

PR #159, opened 2026-05-01
fix: protect job frontmatter from Claude overwrites + auto-recover corrupted sessions

TerrysPOV:
## Summary

Two related reliability fixes for scheduled job runs:

### 1. Job frontmatter protection (`snapshotJobFrontmatter`)

Ports #25 by @Aleschka215.

Scheduled jobs call Claude with the job prompt, but Claude can accidentally overwrite the job file's YAML frontmatter (schedule, model, timeout, etc.), breaking future scheduling. This fix:

- Snapshots the frontmatter before each job run
- After the run completes, checks if frontmatter changed
- Restores original frontmatter if it was modified (preserves Claude's edits to the prompt body)

### 2. Corrupted session auto-recovery (`SIGNATURE_ERROR`)

Ports #26 by @Aleschka215 (combined with #25 — both address job reliability).

When a session is corrupted (Claude returns `Invalid * signature * thinking block`), claudeclaw now:

- Detects the error via `SIGNATURE_ERROR` regex
- Backs up the corrupted session via `backupSession()`
- Retries the run fresh (without `--resume`) automatically

## Files changed

- `src/jobs.ts` — adds `snapshotJobFrontmatter()` exported function
- `src/commands/start.ts` — wraps `runJob()` body with snapshot/restore
- `src/runner.ts` — adds `SIGNATURE_ERROR` detection + fresh-retry block

## Attribution

Ports #25 and #26 by @Aleschka215. Combined since they address the same reliability surface and share the same job-lifecycle code path.

## Validation

- [x] `bunx tsc --noEmit` passes
- [x] `bun run bump:plugin-version && bun run bump:marketplace-version` run

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.21",
+      "version": "1.0.22",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.20",
+      "version": "1.0.21",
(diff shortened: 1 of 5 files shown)


TerrysPOV:
Cherry-picked into #164 with attribution preserved. Thank you for the work on this contribution.

PR #160, opened 2026-05-01
feat(stt): opt-in delegateTool to replace whisper with any transcription tool

TerrysPOV:
## Summary

- Adds `stt.delegateTool` config field (optional string)
- When set, whisper is skipped; Claude is instructed to call that specific MCP tool or CLI command with the audio file path
- When unset (default), the existing whisper transcription path is unchanged — no behaviour change for existing users

## Motivation

Ports the intent of PR #35 (originally titled as STT delegation). The PR was closed with feedback that the implementation was wrong — it stripped whisper entirely and used a vague "use any available transcription tool" prompt. The correct design is an explicit opt-in: the user names the tool they want, Claude calls it by name.

Example config:
```json
{
  "stt": {
    "delegateTool": "mcp__whisper__transcribe"
  }
}
```
Or for a CLI tool: `"delegateTool": "faster-whisper"`.

## Validation

- [ ] `bunx tsc --noEmit` — passes
- [ ] Default (no `delegateTool`): voice messages transcribed via whisper as before
- [ ] With `delegateTool` set: whisper skipped, Claude receives file path + instruction to call named tool
- [ ] Download failure: "downloading failed" message unchanged
- [ ] Transcription failure (whisper path): "could not be transcribed" fallback message

## Plugin version bumps

```
bun run bump:plugin-version   # ✓ → 1.0.22
bun run bump:marketplace-version  # ✓ → 1.0.22
```

Ported from: PR #35

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.20",
+      "version": "1.0.22",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.20",
+  "version": "1.0.22",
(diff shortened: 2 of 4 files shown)


TerrysPOV:
Cherry-picked into #164 with attribution preserved. Thank you for the work on this contribution.

PR #162, opened 2026-05-02
feat: encrypted secrets vault for credential management

Nibbler1250:
## Summary

Adds a Python module for **Fernet-encrypted credential storage**. Agents that need API keys, tokens, or passwords can store them encrypted on disk instead of plaintext in config files.

### Features
- Encrypt/decrypt secrets with a Fernet master key
- Store in `~/.claudeclaw/vault/` (outside repo, not committed)
- CLI interface: `vault.py set/get/list/delete`
- Auto-creates vault directory with restricted permissions (700)
- Master key generated on first use, stored separately

### Files
- `examples/python/vault.py` — standalone module, only requires `cryptography` package
- `examples/python/README.md` — usage documentation

### Usage
```python
from vault import SecretsVault

vault = SecretsVault()
vault.set("TELEGRAM_BOT_TOKEN", "123456:ABC-DEF...")
vault.set("OPENAI_API_KEY", "sk-...")

token = vault.get("TELEGRAM_BOT_TOKEN")
vault.list()  # shows key names (not values)
```

### CLI
```bash
python3 vault.py set TELEGRAM_BOT_TOKEN "123456:ABC..."
python3 vault.py get TELEGRAM_BOT_TOKEN
python3 vault.py list
python3 vault.py delete TELEGRAM_BOT_TOKEN
```

### Context
Running an autonomous agent 24/7, I kept finding tokens in plaintext across `.env` files, config files, and shell scripts. This vault centralizes them encrypted. Built for my agent Greg but designed to be reusable.

No changes to core ClaudeClaw code — purely additive in `examples/python/`.

### DIFF
--- a/examples/python/README.md
+++ b/examples/python/README.md
@@ -1,0 +1,51 @@
+# vault.py - Encrypted Secrets Vault
+
+Fernet-encrypted (AES-128-CBC + HMAC-SHA256) secrets vault for ClaudeClaw
+agents. Credentials are never stored in plaintext on disk.
+
+Requires: `pip install cryptography`
+
+## Features
+
+- Fernet symmetric encryption (AES-128-CBC + HMAC-SHA256)
+- Auto-generates vault key on first use (chmod 600)
+- Key source priority: env var > key file > auto-generate
+- Fallback to environment variables for missing keys
+- Import secrets from .env files
+
+## Usage
+
+```python
+from vault import vault
+
+# Read a secret (falls back to env var if not in vault)
+token = vault.get("TELEGRAM_BOT_TOKEN")
+
+# Store a secret (encrypted on disk)
+vault.set("API_KEY", "sk-...")
+
+# List stored keys (values stay encrypted)
+print(vault.list_keys())
+
+# Bulk import from .env
+vault.import_from_env_file(".env")
+```
+
+## Key Management
+
+The vault key is stored at `~/.claudeclaw/vault/.vault_key` (chmod 600).
+Override with the `CLAUDECLAW_VAULT_KEY` environment variable.
+
+## CLI Usage
+
+The same module is also a CLI:
+
+```bash
+python3 vault.py set TELEGRAM_BOT_TOKEN "1234:abc..."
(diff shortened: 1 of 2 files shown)


TerrysPOV (changes requested):
Thanks for the contribution. The PR currently advertises a vault.py set/get/list/delete CLI, but the file has no CLI entrypoint, so python3 examples/python/vault.py list exits successfully without doing anything. Please either add the CLI handler or update the PR/docs so the advertised usage matches the implementation.

Please note there is a new Master so this will need to rebased to this.

Nibbler1250:
Hey @TerrysPOV — pushed the fix for the missing CLI you flagged.

**Changes (commit `fc5b2ee`):**
- Added an `argparse` CLI handler at the bottom of `examples/python/vault.py` covering the full advertised surface: `set`, `get`, `list`, `delete`, `import`.
- `get` exits 1 on missing key (so it composes well in shell scripts); `list` prints sorted keys (or `(vault empty)`); `import` supports `--overwrite`.
- Updated `examples/python/README.md` to document the CLI alongside the existing Python API examples, so the docs match the implementation.

Smoke-tested all five commands against a fresh `$HOME` (vault key auto-generates, set/get/delete round-trip, list shows correct state, import reads a .env).

Also rebased onto current master. Ready for another look when you have a moment.

TerrysPOV:
I have pushed a small maintainer follow-up commit to address the review feedback: vault decrypt/key failures now fail closed so an existing encrypted vault cannot be overwritten after a bad key or corrupt file read, and the remaining comments and CLI text are in English. Local validation passed: python3 -m py_compile, git diff --check, CLI help, and a smoke test covering set/get/delete plus wrong-key no-overwrite behaviour. This is now merge-ready from my side and will be included and closed via the next bulk cherry-pick PR. Thanks again for the contribution.

TerrysPOV:
This has now been cherry-picked into the bulk merge PR #172, including the maintainer follow-up fix to fail closed on vault decrypt errors. I am closing this source PR to avoid duplicate merge history.

Thank you @Nibbler1250 for the contribution.

PR #163, opened 2026-05-02
feat(discord): isolated sessions per guild channel

BCDel89:
## Summary
- Each guild channel (and thread) now gets its own isolated session keyed by `channelId`; DMs continue to share the global session.
- Adds optional `discord.channelNames` config (channelId → friendly name) to surface a human-readable channel label in the prompt prefix (`[Discord Channel: <name>]` for guild messages, `[Discord DM]` for DMs).

## Test plan
- [ ] Send a message in two different guild channels and verify each gets its own session context.
- [ ] Send a message in a thread and verify it stays isolated under the thread's channelId.
- [ ] Send a DM and verify it still uses the global session.
- [ ] Configure `discord.channelNames` and confirm the friendly name appears in the prompt prefix; absent config falls back to channelId.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.24",
+      "version": "1.0.25",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.24",
+  "version": "1.0.25",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/__tests__/jobs.test.ts
+++ b/src/__tests__/jobs.test.ts
@@ -166,6 +166,18 @@ describe("sessions — agent-scoped paths", () => {
     // Verify path uses getAgentsDir() (project root) not HEARTBEAT_DIR (.claude/...)
     expect(src).toContain('join(getAgentsDir(), agentName, "session.json")');
   });
+
+  test("fallback sessions can be scoped by thread id", async () => {
+    const sessionsSrc = await Bun.file(join(import.meta.dir, "../sessions.ts")).text();
+    const runnerSrc = await Bun.file(join(import.meta.dir, "../runner.ts")).text();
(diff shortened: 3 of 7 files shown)


TerrysPOV (changes requested):
Thanks for the contribution. I found one blocking issue: normal guild-channel messages are routed through a channel-specific session key, but slash commands still use the global session unless the channel is in knownThreads. As a result, /reset, /compact, /status, and similar commands can target the wrong session in ordinary guild channels. Please apply the same guild channel session key resolution to the command handlers.

Please note there is a new Master so this will need to rebased to this.

TerrysPOV (changes requested):
Found 1 issue:                               
                                                                                                                                                                                                                                              
  Before this PR, channels in listenChannels shared the global session — so users there had a continuous conversation history. After this PR, all guild channels (including listenChannels) get routed to per-channel sessions. Any existing conversation history in the global session quietly disappears the moment the daemon restarts. No warning, no log line, nothing.
                                                                                                                                                                                                                                              
  The fix doesn't need to be a full migration — just make the transition visible. A simple warning on the first message that routes to a new per-channel session would do it:                                                                 
                                               
  const sessionKey = isGuild ? channelId : undefined;                                                                                                                                                                                         
  // Warn once when a guild channel transitions from global session to a per-channel one                                                                                                                                                    
  if (sessionKey) {                                                                                                                                                                                                                           
    const existing = await peekThreadSession(sessionKey);
    const globalSession = await peekSession();                                                                                                                                                                                                
    if (!existing && globalSession) {                                                                                                                                                                                                         
      console.warn(`[Discord] Channel ${channelId} now using isolated session. Global session history is no longer accessible here.`);
    }                                                                                                                                                                                                                                         
  }                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                              
https://github.com/moazbuilds/claudeclaw/blob/b909487281739bd65e99fef96ca42448c6a2d0e0/src/commands/discord.ts#L697-L703

TerrysPOV:
I have pushed a small maintainer follow-up commit to address the review feedback: fallback sessions are now scoped by the same thread/channel key used for isolated guild-channel primary sessions, and guild-channel /reset clears the matching fallback state. I also added a regression assertion covering the thread-scoped fallback wiring. Local validation passed: git diff --check, bun install, bun test, and bunx tsc --noEmit --pretty false. This is now merge-ready from my side and will be included and closed via the next bulk cherry-pick PR. Thanks again for the contribution.

TerrysPOV:
This has now been cherry-picked into the bulk merge PR #172, including the maintainer follow-up fix for thread/channel-scoped fallback sessions. I am closing this source PR to avoid duplicate merge history.

Thank you @BCDel89 for the contribution.

PR #165, opened 2026-05-02
fix(start,telegram): pause heartbeats/jobs/messages when rate-limited

TerrysPOV:
## Summary

When Claude hits a usage limit, the daemon was continuing to fire heartbeats and cron jobs, each generating a failed notification — contributing to notification spam and wasted resource on constrained servers.

- **Heartbeat `tick()`**: checks `isRateLimited()` first; if limited, logs the skip and sends a single one-time notification to Telegram and Discord with the reset time
- **Cron `setInterval`**: returns immediately while rate-limited (skips all scheduled jobs)
- **Telegram `handleMessage`**: replies immediately with a human-readable reset time instead of calling Claude

Auto-resumes when `isRateLimited()` clears (the rate limit window expires).

The rate limit state machine in `runner.ts` was already in master. This PR wires the existing exports (`isRateLimited`, `getRateLimitResetAt`, `wasRateLimitNotified`, `markRateLimitNotified`) into the heartbeat and Telegram paths.

## Validation

- [ ] `bunx tsc --noEmit` — passes
- [ ] Rate limit detected → single notification sent, subsequent heartbeats log "skipped"
- [ ] Cron jobs skip cleanly while limited
- [ ] Telegram messages receive immediate reset-time reply
- [ ] After reset time passes → normal operation resumes

## Plugin version bumps

```
bun run bump:plugin-version   # ✓
bun run bump:marketplace-version  # ✓
```

Ported from: PR #33 by @dmitryanchikov (Dmitry Anchikov)

PR #168, opened 2026-05-02
fix(runner): auto-recover from stale --resume session IDs

TerrysPOV:
## Summary

When Claude Code no longer has a session on disk (cleared, expired, compacted away, or synced to another machine), every `--resume` call fails with:

```
Error (exit 1): No conversation found with session ID: <id>
```

The daemon would loop forever returning that error — the only recovery was to hand-delete `session.json`. Telegram and Discord users saw a permanent error with no way out.

**Fix:** both `execClaude` and `streamClaude` now detect the error, back up the dead session ID, and retry as a fresh session — all within one turn.

### Changes

- `STALE_SESSION_PATTERN` + `isStaleSessionError()` — detects the error in stdout/stderr
- `stripResume()` — removes `--resume <id>` from argv before retry
- `withOutputFormat()` — forces `stream-json` on retry to capture the new session ID
- `execClaude` recovery: removes thread session or backs up global, retries once, persists new session ID
- `streamClaude` recovery: collects stderr in parallel (avoids backpressure), detects stale after exit, backs up and recurses once
- Auto-compact-on-timeout and turn-count increment gated with `!recoveredFromStale`

### Repro

```bash
echo '{"sessionId":"deadbeef-dead-dead-dead-deaddeaddead","createdAt":"...","lastUsedAt":"...","turnCount":5}' \
  > .claude/claudeclaw/session.json
# Any Telegram message → "Error (exit 1): No conversation found..."
# Every subsequent message → same error
```

After this fix: logs `Stale session deadbeef for <name>; recovering...`, then returns a normal reply.

## Validation

- [ ] `bunx tsc --noEmit` — passes
- [ ] Injected stale ID → single recovery log line + successful reply
- [ ] Normal path unchanged (no stale session → no extra code executed)
- [ ] Thread session: `removeThreadSession` used instead of `backupSession`

## Plugin version bumps

```
bun run bump:plugin-version   # ✓ → 1.0.24
bun run bump:marketplace-version  # ✓ → 1.0.24
```

Ported from: PR #110 by @lunklin

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.23",
+      "version": "1.0.24",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.23",
+  "version": "1.0.24",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -959,6 +959,8 @@ async function execClaude(
 
     if (threadId) {
       await removeThreadSession(threadId);
+    } else if (agentName) {
+      await resetSession(agentName);
     } else {
       await backupSession();
     }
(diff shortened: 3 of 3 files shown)


TerrysPOV:
@codex review

TerrysPOV:
@codex review

TerrysPOV:
@codex review

TerrysPOV:
Closing this in favour of the bulk cherry-pick PR #169. The stale --resume session recovery work from this PR is the source of truth used in the bulk branch, which avoids duplicating the overlapping stale-session recovery commits from #165 while preserving attribution in the cherry-picked commit history and PR body. Thanks for the contribution.

PR #170, opened 2026-05-02

TerrysPOV:
## What

Guard `alive()` against an empty or corrupt pidfile causing `process.kill(0, 0)` to silently report the daemon as live.

`process.kill(0, signal)` in Node operates on the *calling process group*, not a specific PID, and always succeeds — so an empty pidfile (daemon crashed mid-write, lifecycle race) returned `true` forever, leaving the statusline stuck on `● live` and blocking daemon restarts without a manual `rm daemon.pid`.

The fix validates the parsed PID is a finite, positive integer before calling `process.kill`. Any other value returns `false` immediately.

## Fix

```js
function alive() {
  try {
    var parsedPid = Number(pid);
      return false;
    }
    process.kill(parsedPid, 0);
    return true;
  } catch { return false; }
}
```

## Test plan

- Empty pidfile → `alive()` returns `false` (was `true`)
- Pidfile contains `"0"` → `false` (was `true`)
- Pidfile contains a valid running PID → `true` (unchanged)
- Pidfile contains a stale PID → `false` (unchanged)

## Attribution

Originally reported and fixed by [@prashantsolanki3](https://github.com/prashantsolanki3) in #127. Ported to a clean branch off current master. Thank you Prashant.


TerrysPOV:
@codex review

TerrysPOV:
Closing — branch was accidentally based off ClaudeClaw-Plus main (207 commits ahead of master). Replacing with a clean single-commit branch.

PR #182, opened 2026-05-03
feat(jobs): persist lastResult and lastRanAt per job in state

### LINKED ISSUE #175: feat(jobs): persist lastResult and lastRanAt per job in state
## Problem

After a daemon crash or restart, jobs lose all context of what they were doing. The next run starts cold with no knowledge of whether the previous run succeeded, what it produced, or when it last ran.

## Proposed solution

Add `lastResult` and `lastRanAt` fields per job in `state.json`:

```json
{
  "jobs": [{
    "name": "morning-brief",
    "nextAt": 1234567890,
    "lastResult": "ok",
    "lastRanAt": 1234567000
  }]
}
```

- `lastResult`: `"ok"` | `"error"` | `"skipped"`
- `lastRanAt`: Unix timestamp of last completion

The heartbeat prompt can reference `lastResult` to avoid re-doing completed work after a restart. No breaking changes — fields are additive.

## Source

Requested by @Nibbler1250 in #14 (Mega-Post).

Nibbler1250:
Resolves #175.

Adds two optional fields to each job entry in `state.json` so the statusline / heartbeat prompt / external monitors can see what each job last did without re-executing it.

## Changes

**`src/statusline.ts`** — extend `StateData.jobs[]` type:
```typescript
jobs: {
  name: string;
  nextAt: number;
  /** Outcome of the most recent run. Absent until the job runs at least once. */
  lastResult?: "ok" | "error" | "skipped";
  /** Unix timestamp (ms) of the most recent completion. Absent until first run. */
  lastRanAt?: number;
}[];
```

**`src/commands/start.ts`** — three small additions:

1. Declare an in-memory `jobLastResult` Map alongside the existing `jobRetryState` Map (same lifecycle: resets on daemon restart, no disk persistence).
2. In the `runJob()` `.then()` handler, right after the run completes (next to where `jobRetryState` is updated), record the outcome:
   ```typescript
   jobLastResult.set(job.name, {
     result: r.exitCode === 0 ? "ok" : "error",
     ranAt: Date.now(),
   });
   ```
3. In `updateState()`, spread the recorded fields into each job entry so they land in `state.json` on the next 60-second tick:
   ```typescript
   jobs: currentJobs.map((job) => {
     const last = jobLastResult.get(job.name);
     return {
       name: job.name,
       nextAt: nextCronMatch(...).getTime(),
       ...(last ? { lastResult: last.result, lastRanAt: last.ranAt } : {}),
     };
   })
   ```

## Properties

- **Additive**: fields are absent until the job runs at least once; existing consumers see no change.
- **No new disk I/O**: piggybacks on the existing `updateState()` 60s tick.
- **No new dependency**: just a `Map<string, { result, ranAt }>` in memory.
- **Lifecycle matches `jobRetryState`**: in-memory only, drops on daemon restart (no stale debt across restarts, consistent with how retry state already works in this file).

## Notes on `"skipped"`

I included `"skipped"` in the union type for forward compatibility (e.g. a future heartbeat could mark a job skipped due to rate-limit / disabled / pause), but **this PR doesn't write `skipped` anywhere yet**. Happy to drop that variant if you'd rather keep the surface tight to `"ok" | "error"` for now.

## Build

- `bun build src/index.ts --target=bun` → 0.42 MB, clean.
- `bunx tsc --noEmit` → no new errors. The two pre-existing errors at `start.ts:442` / `:447` (`forwardToTelegram` / `forwardToDiscord` arity mismatch) live on `master` already; I left them alone.
- No new lockfile / package changes.

## Source

Originally requested by myself in #14 (Mega-Post) — thanks @TerrysPOV for breaking it out into its own ticket.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.27",
+      "version": "1.0.28",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.27",
+  "version": "1.0.28",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -750,10 +750,14 @@ export async function start(args: string[] = []) {
       heartbeat: currentSettings.heartbeat.enabled
         ? { nextAt: nextHeartbeatAt }
         : undefined,
-      jobs: currentJobs.map((job) => ({
-        name: job.name,
-        nextAt: nextCronMatch(job.schedule, now, currentSettings.timezoneOffsetMinutes).getTime(),
-      })),
+      jobs: currentJobs.map((job) => {
+        const last = jobLastResult.get(job.name);
+        return {
+          name: job.name,
+          nextAt: nextCronMatch(job.schedule, now, currentSettings.timezoneOffsetMinutes).getTime(),
+          ...(last ? { lastResult: last.result, lastRanAt: last.ranAt } : {}),
+        };
+      }),
       security: currentSettings.security.level,
       telegram: !!currentSettings.telegram.token,
       discord: !!currentSettings.discord.token,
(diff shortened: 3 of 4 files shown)


TerrysPOV (changes requested):
Found 2 issues:

  1. Daemon crashes at startup when any jobs are configured — updateState() is called immediately at line 770, but const jobLastResult (which the updated updateState() body now reads via jobLastResult.get(job.name)) is declared at line ~779. const variables are in the temporal dead zone until their declaration is reached; accessing one before that throws ReferenceError: Cannot access 'jobLastResult' before initialization. Any user with jobs configured will see the daemon die on startup. Fix: move the jobLastResult declaration above the updateState() call (alongside or before jobRetryState).

  https://github.com/moazbuilds/claudeclaw/blob/a7dd0d06ee13b46d6be997179eabe4c5777045e0/src/commands/start.ts#L750-L757

  https://github.com/moazbuilds/claudeclaw/blob/a7dd0d06ee13b46d6be997179eabe4c5777045e0/src/commands/start.ts#L776-L780

  2. "skipped" is declared in the type but never written — The type union in statusline.ts includes "skipped", and jobLastResult is typed to accept it, but the only write path sets "ok" or "error" from r.exitCode. The natural trigger for "skipped" — jobs bypassed because isRateLimited() returns true — has no jobLastResult.set() call. Any consumer checking lastResult === "skipped" to detect rate-limit suspensions will never see it. Fix: set { result: "skipped", ranAt: Date.now() } inside the isRateLimited() branch of the setInterval, once per job that was due to fire.

  https://github.com/moazbuilds/claudeclaw/blob/a7dd0d06ee13b46d6be997179eabe4c5777045e0/src/statusline.ts#L8-L12

Please effect these changes, commit and resubmit.

TerrysPOV (approved):
This is now good to merge.

PR #184, opened 2026-05-03
feat(telegram): live streaming, /verbose tool display, /fork, /kill

TerrysPOV:
## Summary

Port of [PR #21](https://github.com/moazbuilds/claudeclaw/pull/21) by @1872-svg, cleanly rebased onto current master.

### What this adds

- **Live streaming responses** — the Telegram handler now sends a `⏳` placeholder immediately, then edits it in real-time as Claude generates text (throttled at 500ms via `editMessageText`). Final message is sent as formatted HTML.

- **`/verbose`** — per-chat toggle. When on, tool call summaries (`● Bash(...)`, `⎿ [Read] done`) appear above the streaming text so the user can see what Claude is doing in real time.

- **`/kill`** — kills the currently running Claude subprocess. Uses a new `killActive()` export in `runner.ts` that tracks the active `Bun.spawn` process.

- **`/fork`** — runs a prompt through a parallel lightweight Haiku agent without touching the main session queue. Useful for quick questions while Claude is working. The main handler also automatically forks when `isMainBusy()` returns true (i.e. a message arrives while Claude is mid-response).

### runner.ts changes

- `activeProc` + `killActive()`: track and kill the active subprocess
- `mainRunning` + `isMainBusy()`: signal when the main serial queue is executing
- `runClaudeStreaming()`: new streaming runner that emits `onChunk`/`onToolEvent` callbacks from NDJSON events — separate from the existing `runClaudeStream()` which is kept intact
- `formatToolCallSummary()`, `extractToolResultText()`: helpers for verbose display
- `runFork()`: spawns a Haiku agent outside the main queue (uses `runClaudeOnce`, not `enqueue`)
- `runClaudeStream()` and `runUserMessage()` extended with optional `onChunk`/`onToolEvent` params — fully backward-compatible (all existing Discord/Slack/job callers unaffected)

### Backward compatibility

All existing callers of `run()`, `runUserMessage()`, `execClaude()`, and `runClaudeStream()` work without changes. The new parameters are optional and default to undefined/no-op.

## Test plan

- [ ] Start the Telegram bot and send a message — verify the `⏳` placeholder appears immediately and edits to the final response
- [ ] Send `/verbose` — verify ack message, then send a message and confirm tool call lines appear above the response text
- [ ] Send `/verbose` again — verify verbose mode turns off
- [ ] While a long Claude task is running, send another message — verify it routes to the fork agent without blocking
- [ ] Send `/fork what is 2+2` — verify fast response from Haiku without blocking the main session
[...]

PR #183, opened 2026-05-03
feat(discord): include reply and forwarded message context in prompt

BCDel89:
## Summary

- Injects replied-to message content into the Claude prompt as `[In reply to <user>]: <text>`
- Injects forwarded message snapshot content as `[Forwarded message from <user>]: <text>`
- Attachments on referenced/forwarded messages are noted by filename but not downloaded
- Adds `DiscordMessageSnapshot` interface and extends `DiscordMessage` with `message_reference` and `message_snapshots` fields

## Test plan

- [ ] Reply to a message in a Discord guild channel — verify the bot's response reflects the replied-to content
- [ ] Forward a message into a channel where the bot is listening — verify the forwarded content appears in the bot's context
- [ ] Send a plain message (no reply, no forward) — verify no regression
- [ ] Reply to a message that has attachments — verify attachment filenames are noted in the prompt

Closes BCD-89

PR #173, opened 2026-05-03
telegram: per-topic session isolation

ybuzko:
## Summary

- Adds `getTelegramSessionKey()` helper that maps each Telegram context to an isolated session key: `tg:<chatId>:<threadId>` for supergroup topics, `tg:<chatId>` for general group chat, `tg:dm:<userId>` for per-user DMs (opt-in), or `undefined` to fall through to the global session
- Adds `telegram.dmIsolation: "shared" | "perUser"` config setting (default `"shared"` to match existing Discord DM behavior)
- Scopes `/reset`, `/compact`, `/status`, and `/context` to the per-topic session when a key is present

## Motivation

Without this, all Telegram group topics share a single global session. A conversation in one topic bleeds context into unrelated topics, and `/reset` nukes every topic at once. This matches the isolation pattern Discord already uses for guild channels.

## Test plan

- [ ] Supergroup: send messages in two different topics — confirm each maintains its own context
- [ ] `/reset` in one topic does not affect another topic's session
- [ ] `/status` and `/context` show the correct per-topic session
- [ ] `/compact` runs on the correct thread session
- [ ] DMs with `dmIsolation: "shared"` (default) still use the global session
- [ ] DMs with `dmIsolation: "perUser"` isolate per user ID
- [ ] Plain group chat (no threads) uses `tg:<chatId>` key correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #185, opened 2026-05-03
feat(security): Phase 1 security hardening — lock the front door

lohengrin332:
## Summary

My apologies that I didn't open an issue for all this first. I have more changes, and will open issues for them before submitting additional CRs.

This changeset is in response to a detailed security review and remediation plan by Claude Opus 4.7, with changes implemented by Claude Sonnet 4.6.

Web dashboard authentication
- Generate a random 256-bit bearer token on first start; store at .claude/claudeclaw/web.token (mode 0600)
- All /api/* routes require the token via Authorization: Bearer header or ?token= query param; /api/inject also accepts the legacy settings.apiToken so existing automation isn't broken
- Dashboard JS reads the token from ?token= on load, stores it in sessionStorage, strips it from the URL, and auto-attaches it to every fetch() call; shows a paste-token prompt if none is found
- Host header validation (DNS-rebinding defense): reject requests whose Host doesn't match the configured bind address; wildcard binds (0.0.0.0, ::) are exempted since browsers send the real hostname there
- Origin validation for POST/DELETE (CSRF defense): cross-origin requests with a set Origin header are rejected with 403
- timingSafeEqual now compares Buffer byte lengths, not JS string character lengths, so non-ASCII tokens can't cause RangeError

Secrets leak fix (/api/technical-info)
- Recursive redact() replaces any field named token, api, apiKey, apiToken, botToken, appToken, password, or secret with "<redacted:N chars>" before the response is sent
- Snapshot uses sanitizeSettings() rather than the raw settings object

Fail-closed allowlists
- Telegram and Discord: deny if allowedUserIds is empty or the sender is not in the list (previously fail-open when list was empty)
- Daemon refuses to start if a bot token is configured but allowedUserIds is empty
- Telegram handleMyChatMember: leaves groups added by unauthorized users instead of posting a welcome message
- Discord handleGuildCreate: stays silent unless the guild ID is in the new allowedGuilds list (new DiscordConfig field, default [])
- Discord interaction handler: same fail-closed check

Prompt-safety wrapper (wrapUntrusted)
- New src/prompt-safety.ts: wraps attacker-controlled text in <untrusted-label-id> blocks, truncates to maxLen, and defangs any matching closing tags so content can't break out of the wrapper
- Applied to: Telegram message text, voice transcripts, attachment filenames, group title/adder username; Discord message text, voice transcripts, text attachments (cap reduced 50 KB → 2 KB), guild name; skill command arguments in both bots
- System prompt in runner.ts instructs Claude to treat <untrusted-...> content as data, not instructions

Test harness
- tests/auth.test.ts: checkToken unit tests including non-ASCII byte-length edge case
- tests/prompt-safety.test.ts: wrap, truncate, and defang tests
- tests/allowlist.test.ts: fail-closed allowlist logic
- tests/integration/curl-suite.sh: curl-based smoke tests for auth, Host validation, Origin validation, and secrets-leak checks
- package.json: add "test": "bun test"
## Validation

- [X] I ran the relevant checks locally
- [X] I updated any docs or setup guidance affected by this change

## Plugin Versioning

If this PR changes shipped plugin files under `src/`, `commands/`, `prompts/`, or `.claude-plugin/`, version metadata may also need to be bumped.

Run as needed:

```bash
bun run bump:plugin-version
```

Typical rule:
- bump `.claude-plugin/plugin.json` when shipped plugin content changes
- bump `.claude-plugin/marketplace.json` when marketplace metadata should reflect the new shipped version

Docs-only and other non-shipped changes do not require these bumps.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.37",
+      "version": "1.1.0",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.1.0",
+      "version": "1.1.1",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.37",
+  "version": "1.1.0",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/README.md
+++ b/README.md
@@ -87,6 +87,24 @@ Prior to this release, an empty `allowedUserIds` list meant **allow everyone**. 
 
 Run `claudeclaw config` for guided setup if you're unsure of your user ID.
 
+### v1.1.0 — Web UI bearer token gate
+
+All `/api/*` routes (except `/api/health`) now require an `Authorization: Bearer <token>` header. The token is auto-generated on first start and written to `.claude/claudeclaw/web.token`. The daemon also prints the full URL with the token embedded when the web UI starts.
+
+**Migration:** update any scripts that call `/api/state` or other API routes to pass the token:
+
+```
+Authorization: Bearer <contents of .claude/claudeclaw/web.token>
+```
+
+Existing `/api/inject` users who configured `settings.apiToken` are unaffected; that fallback still works.
+
+### v1.1.0 — Discord text-attachment truncation limit reduced
+
+Text attachments sent to the Discord bot are now truncated at **2,048 bytes** (previously 51,200). Payloads over that limit have `…[truncated]` appended silently; there is no config knob to restore the old limit.
+
+**Migration:** if you rely on passing large text files through Discord attachments, switch to gists or another file-sharing mechanism and paste the URL instead.
+
 ---
 
 ## What Would Be Built Next?
--- a/README.md
+++ b/README.md
@@ -70,6 +70,25 @@ bun run bump:marketplace-version
 
 Docs-only and other non-shipped changes do not require these bumps.
 
+## Upgrading
+
+### v1.0.26 — Allowlist behavior change (Telegram & Discord)
+
+Prior to this release, an empty `allowedUserIds` list meant **allow everyone**. That was a potential security vulnerability; any Telegram or Discord user could drive the daemon.
+
+**New behavior:** an empty list means **block everyone**. The daemon will refuse to start if a bot token is configured without at least one allowed user ID.
+
+**Migration:** add your user ID(s) to `settings.json` before upgrading:
+
+```json
+"telegram": { "allowedUserIds": [123456789] },
+"discord":  { "allowedUserIds": ["987654321012345678"] }
+```
+
+Run `claudeclaw config` for guided setup if you're unsure of your user ID.
+
+---
+
 ## What Would Be Built Next?
 
 > **Mega Post:** Help shape the next ClaudeClaw features.
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "telegram": "bun run src/index.ts telegram",
     "discord": "bun run src/index.ts discord",
     "status": "bun run src/index.ts status",
+    "test": "bun test",
     "bump:plugin-version": "bun run scripts/bump-plugin-version.ts",
     "bump:marketplace-version": "bun run scripts/bump-marketplace-version.ts"
   },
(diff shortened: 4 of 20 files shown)


TerrysPOV:
COMPLEX — 18 files, security-critical changes across server, Discord, Telegram, and prompt handling subsystems. Triage before merge.

Vision: ALIGNED — security hardening is firmly within claudeclaw's scope and doesn't add heavyweight dependencies.

Overlapping PRs: #184 touches `telegram.ts` and `runner.ts`; #183 touches `discord.ts`; #182 touches `start.ts`; #173 touches `telegram.ts` and `config.ts`. All four will have merge conflicts with this PR if it lands first.

### Code review

Found 4 issues:

**1. CSRF origin check always rejects HTTPS**

`server.ts` builds `expectedOrigin` as `http://${host}` unconditionally. Any browser connecting over HTTPS sends `Origin: https://...`, which never matches, so every mutating request from an HTTPS deployment gets a 403. Fix: derive the scheme from the request (`req.url.startsWith("https://") ? "https" : "http"`) or accept both schemes for the same host.

**2. `allowedUserIds` logic inversion is a breaking change**

`discord.ts` and `telegram.ts` previously blocked requests when `allowedUserIds.length > 0 && !allowedUserIds.includes(userId)` — meaning an empty list was "allow all". The PR flips this to `allowedUserIds.length === 0 || !allowedUserIds.includes(userId)` — an empty list now blocks everyone. Existing deployments that haven't set `allowedUserIds` will silently stop working after upgrade with no error message. Either restore the original semantic and require an explicit opt-in flag (`requireAllowList: true`) to get the new stricter behaviour, or add a startup warning and document the migration in the README.

**3. `wrapUntrusted` doesn't cover opening tags or cross-label closing tags**

`prompt-safety.ts` strips `</untrusted-user-message-${id}>` but leaves opening tags (`<untrusted-user-message-ANYID>`) untouched. An attacker can inject an opening tag for a *different* ID, or inject a closing tag for the *outer* label, both of which pass through unmodified and can break the structure the model sees. The stripping needs to cover `<untrusted-user-message-[^>]*>` (any opening variant) and `</untrusted-user-message-[^>]*>` (any closing variant), not just the exact label for this message.

**4. `/api/health` now returns 401**

The bearer-token gate at the top of the fetch handler matches `url.pathname.startsWith("/api/")`, and the `/api/health` handler sits below it. Monitoring tools, load balancers, and uptime checks that hit `/api/health` unauthenticated will all start getting 401s after this lands. Move the `/api/health` early-return to above the auth gate (same pattern as `/api/csrf-token` which is also pre-auth).

TerrysPOV (changes requested):
Thanks for the security hardening work here. I rechecked the current branch and I’m not including this in the #209 bulk cherry-pick because it needs separate cleanup first.

The branch is currently conflicting against `master`, and `git diff --check` shows committed conflict markers in `.claude-plugin/marketplace.json` and `src/commands/telegram.ts`. Given this PR’s security-sensitive scope across auth, prompt handling, Telegram, Discord, runner, and UI code, it should be cleaned up/rebased and reviewed as its own security PR rather than folded into a feature bulk merge.

lohengrin332:
I'll try to rebase and clean it up this weekend. Thanks.

-JD

lohengrin332:
Rebase is done and tests pass locally. I did not bump the version numbers because I thought you might want to bump a minor version (e.g. 1.1.0), given that these are breaking changes in a small way. LMK if you would like me to make this change or any other. Thanks.

TerrysPOV (changes requested):
Thanks JD, clean rebase and the four issues from last round are properly fixed. Nice call on the startup hard-exit for the empty allowlist case — that's exactly the right shape for the migration.

You're right that this warrants a minor bump rather than patch. Two asks before I approve:

**1. Bump to 1.1.0 in both plugin files.** `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json`. The `bun run bump:*` scripts do a patch bump, so you'll need to set those values by hand.

**2. Extend the README "Upgrading" section to cover the other two breaking changes.** Right now it only documents the allowlist flip, but operators will also be caught by:

- **Web UI bearer token gate.** Any script hitting `/api/state` or other `/api/*` routes (except `/api/health`) now needs `Authorization: Bearer <token>`. The token is auto-generated and written to `.claude/claudeclaw/web.token` on first start. Existing `/api/inject` users on `settings.apiToken` are fine — that fallback still works.

- **Discord text-attachment truncation.** Limit dropped from 51,200 to 2,048 bytes. Larger attachments now get `…[truncated]` appended silently and there's no config knob to restore the old size. Worth a sentence so operators know to switch to gists or file uploads for larger payloads if needed.

Once those two land I'll approve.

Two optional follow-up nits you can address here or punt to a separate PR — your call:

- `tests/allowlist.test.ts` defines its own `isAllowed()` stub rather than importing the real guard from `telegram.ts` / `discord.ts`. Means the test won't catch logic drift in the production code. Worth wiring up to the real function.
- `wrapUntrusted` regex `[a-z0-9]+` only matches the format your own ID generator uses. If you wanted defence-in-depth against an attacker injecting uppercase or hyphen variants the LLM might still interpret as boundaries, broaden the class. Low priority.

Vision: aligned. Complexity: complex — 17 files, security-sensitive — but the scope is coherent and keeping it as one PR is the right call.

lohengrin332:
I implemented all four changes you requested. Please let me know if you need any further changes before this can be merged. Once this is merged I'll (eventually) start working on my the other issues I reported. Thanks!

TerrysPOV (approved):
Thanks JD, all four follow-ups landed cleanly:

- Versions at 1.1.0 across both plugin files ✓
- README "Upgrading" section now covers all three breaking changes (allowlist flip, web UI bearer token gate, Discord attachment truncation) ✓
- `isAllowed` extracted to `src/allowlist.ts`, both `discord.ts` and `telegram.ts` import the shared function, and the test imports the real fn rather than a stub ✓
- `wrapUntrusted` regex broadened to `[a-zA-Z0-9_-]+` — covers uppercase, underscore, hyphen, still anchored so no false positives ✓

I pushed one tiny maintainer commit on top (`23318aa`) to flip the `"Allowed users: all"` startup banner string to `"none (deny all)"` in `discord.ts` and `telegram.ts`. After the v1.1.0 semantic flip an empty list now means deny-all, so the old label was lying to operators reading the startup log. Bumped versions to 1.1.1 in the same commit.

On the `isAllowed<T>` generic I noticed in the prior round — I left it alone after looking again. It's actually load-bearing because Discord IDs are `string` and Telegram IDs are `number`; the generic couples the userId type to the array element type so callers can't mix them. Removing it would weaken type safety, not improve it. My nit was wrong, apologies for the noise.

Approving.

PR #191, opened 2026-05-05
feat: proactive session resume after self-restart

squirblej:
## Problem

Agents that manage their own lifecycle — applying config changes, deploying updated code, reloading service files — restart their host process with something like:

```bash
nohup bash -c 'sleep 2 && sudo systemctl restart my-service' &
```

The session exits cleanly, but there's no way for the new instance to know what to report back. The user gets silence, or at best a pre-written canned message that was composed *before* the restart and can't reflect what actually happened.

A static message queue patches over this, but has a fundamental flaw: "Restarted successfully" written before the restart fires unconditionally whether the restart succeeded or not — worse than no message when things go wrong.

## Solution

`pending-resume.json`: before triggering a restart, the agent writes a file containing a transport, channel ID, session key, and wake-up prompt. On the next startup, after the gateway reports ready, ClaudeClaw resumes the stored session and injects the wake-up prompt as the next user turn.

The resumed Claude session runs normally — full tool access, original context, ability to check `systemctl status`, read logs, verify config. Its output is delivered to the stored channel and thread.

The distinction matters: the agent isn't pre-writing a reply, it's pre-scheduling a **verification step**. The confirmation the user sees is composed after the restart, by a Claude instance that can actually observe the post-restart state.

ClaudeClaw already has all the session resume infrastructure needed (`--resume`, `sessions.json`, `sessionManager`). The only missing piece was a startup hook that fires proactively without waiting for the next user message.

## Usage

Write `.claude/claudeclaw/pending-resume.json` before triggering the restart:

```json
{
  "transport": "discord",
  "channelId": "1234567890",
  "sessionKey": "optional-thread-session-key",
  "wakeUpPrompt": "The service was just restarted. Check systemctl status and recent logs, then report back.",
  "expires": 1746100000000
}
```

The file is renamed atomically before parsing (prevents double-fire on crash). A module-level `consumed` guard ensures it fires at most once per process lifetime.

## Changes

- `src/pending-resume.ts` — new: `PendingResume` interface and `loadPendingResume()` (atomic rename, expiry check, field validation)
- `src/commands/discord.ts` — `runPendingResume()` called in `READY` handler
- `src/commands/telegram.ts` — `runPendingResumeTelegram()` called in `startPolling`
- `RESTART_RESUME_SPEC.md` — spec describing the design and rationale

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.29",
+      "version": "1.0.30",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.28",
+      "version": "1.0.29",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.29",
+  "version": "1.0.30",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.28",
+  "version": "1.0.29",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/RESTART_RESUME_SPEC.md
+++ b/RESTART_RESUME_SPEC.md
@@ -13,7 +13,7 @@ This kills the systemd service — which is the process running the CC session t
 - The Bash tool receives exit code 144 (process killed mid-execution)
 - The CC session is terminated before it can send a reply
 - The agent cannot confirm success to the user
-- If the session had a pending Discord/Slack/Telegram reply queued, it is lost
+- If the session had a pending Discord/Telegram reply queued, it is lost
 - If the session is resumed later (e.g. user sends another message), CC resumes the old context and may re-run the restart command, causing a second unnecessary restart
 
 This is a fundamental gap: agents that manage their own lifecycle have no clean way to restart and report back.
@@ -52,7 +52,7 @@ ClaudeClaw provides a file-based mechanism for agents to schedule a **session co
 
 2. Agent fires the detached restart and ends the session cleanly.
 
-3. On the next startup, after the transport gateway reports ready (Discord READY event, Slack `hello` + `auth.test`, Telegram polling start), ClaudeClaw atomically renames the file (preventing double-fire on crash) and calls `runUserMessage` with the stored session key and wake-up prompt.
+3. On the next startup, after the transport gateway reports ready (Discord READY event, Telegram polling start), ClaudeClaw reads the file to check the transport, then atomically renames it (preventing double-fire on crash) and calls `runUserMessage` with the stored session key and wake-up prompt.
 
 4. The resumed Claude session runs as a normal turn — it has all its original context, can run tools, read logs, check process state — and its output is delivered to the stored channel and thread.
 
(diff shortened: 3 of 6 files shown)


TerrysPOV:
Vision: PARTIAL — the feature solves a real gap (post-restart confirmation) but a startup hook that fires proactively without user interaction sits uneasily against the repo's "no hidden implicit behaviors" principle.

Overlapping PRs: #185 touches both `discord.ts` and `telegram.ts` and will conflict here; #194 touches the READY handler area of `discord.ts`.

### Code review

Found 3 issues:

**1. `consumed` flag is set before the transport is checked — wrong-transport caller silently destroys the file**

`loadPendingResume()` sets `consumed = true` and renames the file before any caller checks `resume.transport`. In a process with both Discord and Telegram configured, Discord's handler fires first (fire-and-forget in READY), sets `consumed = true`, renames + deletes the file, then returns early because `transport !== "discord"`. Telegram's handler then calls `loadPendingResume()`, gets `null` immediately (consumed flag is set), and the Telegram resume is silently lost — no warning, no log. Fix: pass the expected transport into `loadPendingResume` as a parameter and do the transport check before consuming the file.

https://github.com/moazbuilds/claudeclaw/blob/ec85108550639dcbd5c892b8654d0f3e55ff4aa6/src/pending-resume.ts#L43-L47

**2. `agentName` is declared and documented but never forwarded to `runUserMessage`**

`PendingResume` defines `agentName?: string` ("agent working-directory name if the session is agent-scoped") and the spec calls it out as explicitly supported. But both transport functions call `runUserMessage` with only three arguments — `agentName` is the fourth. Any agent that writes `agentName` into `pending-resume.json` expecting an agent-scoped resume will silently get the global session instead.

https://github.com/moazbuilds/claudeclaw/blob/ec85108550639dcbd5c892b8654d0f3e55ff4aa6/src/commands/discord.ts#L979-L983

https://github.com/moazbuilds/claudeclaw/blob/ec85108550639dcbd5c892b8654d0f3e55ff4aa6/src/commands/telegram.ts#L1044-L1048

**3. `"slack"` is in the transport union type but has no implementation**

The type allows `"discord" | "telegram" | "slack"` and the spec describes Slack support. But there is no `runPendingResumeSlack`. A Slack `pending-resume.json` is atomically consumed (file renamed, then deleted, `consumed = true` set) and then silently dropped — no error is logged. Either remove `"slack"` from the union until it's implemented, or add a `console.warn` when a Slack resume is found and discarded.

https://github.com/moazbuilds/claudeclaw/blob/ec85108550639dcbd5c892b8654d0f3e55ff4aa6/src/pending-resume.ts#L11-L14

TerrysPOV:
Please also rebase this to the current Master.

squirblej:
Thanks for the thorough review — all three bugs are valid and fixed in the latest push:

1. `loadPendingResume` now takes `expectedTransport` and checks it **before** renaming the file. Wrong-transport callers return `null` without touching the file.
2. `agentName` is forwarded to `runUserMessage` in both Discord and Telegram handlers.
3. `"slack"` removed from the union type until there's an actual implementation.

---

On the **implicit behavior** concern: I'd push back gently here. The startup hook fires only when an agent has explicitly written `pending-resume.json` before exiting — there's no ClaudeClaw-side trigger, no config option that silently activates it, no behavior the user didn't request. The agent is making a contract ("I will report back after I restart"), and the startup hook is what honors that contract.

The *implicit* behavior is actually the status quo — an agent that commits to restarting and reporting back just goes silent, and the user has no idea whether the task completed. That's the surprising outcome. The `pending-resume` mechanism is precisely what makes agent restarts explicit and observable.

Framing it as "proactive without user interaction" applies equally to heartbeats, job scheduling, and any other ClaudeClaw feature — all of them act on the user's behalf without a real-time trigger. The opt-in here is the agent writing the file.

Happy to add a note in the README clarifying the mechanism if that would help address the concern.

TerrysPOV (changes requested):
Thanks for the update. I rereviewed the current head (`e9daf1e`) against current master.

The three previous code findings are addressed: `loadPendingResume()` checks transport before consuming, `agentName` is forwarded for Discord and Telegram, and Slack was removed from the runtime transport union. The branch is also now mergeable against current master.

I found two follow-ups before merge:

1. Telegram pending-resume failure can prevent polling from starting. `startPolling()` awaits `runPendingResumeTelegram()` before calling `poll(gen)`, so if the resume run or final send throws, Telegram logs a fatal error and never starts polling. Please isolate pending-resume errors from transport startup, e.g. wrap that call in its own try/catch or fire/log it independently.

2. `RESTART_RESUME_SPEC.md` still documents Slack support and says the file is renamed before parsing. The runtime now supports only Discord/Telegram and parses before rename to check transport, so the spec should be updated to avoid agents writing unsupported Slack pending-resume files.

Local validation passed on the previous head with the same affected code paths:
- `bun install`
- `bun test`
- `bunx tsc --noEmit --pretty false`

The failing `claude-review` check is the known CI credential issue (`ANTHROPIC_API_KEY` empty/invalid).

squirblej:
Both follow-ups addressed in the latest push (f0faf33):

1. `runPendingResumeTelegram()` is now fire-and-log — wrapped in `.catch()` so a resume failure can't block polling startup.
2. `RESTART_RESUME_SPEC.md` updated: Slack removed from the transport table and scope section, parse-before-rename ordering corrected, Telegram snippet reflects the actual error-isolation pattern.

Ready for re-review.

TerrysPOV (approved):
Rereviewed the latest head (`f0faf33`) against current master.

  The previous requested changes are addressed:
  - Telegram pending-resume failures are now isolated from polling startup.
  - `RESTART_RESUME_SPEC.md` now matches the implemented Discord/Telegram-only behaviour and parse-before-rename flow.

  Local validation passed:
  - `bun install`
  - `bun test`
  - `bunx tsc --noEmit --pretty false`

Plugin and marketplace version guards are passing. The remaining `claude-review` failure is the known CI credential issue (`ANTHROPIC_API_KEY` empty/invalid), not a code failure.

No remaining findings from me.

TerrysPOV:
Closing this as the work has been cherry-picked into #196, with the original author attribution preserved in the combined PR. Thanks @squirblej for the contribution.

PR #192, opened 2026-05-05
feat: track concurrent session count in active-runs file

squirblej:
## Problem

Agents restarting their own host process risk interrupting other concurrently-running sessions (e.g. a long task in a thread). There was no way for an agent to check whether other sessions were active before deciding to restart.

## Solution

Writes the number of currently-running main-queue sessions to `.claude/claudeclaw/active-runs` (a plain integer file) on every session start and finish, using `writeFileSync` so writes are synchronous and always reflect the true in-memory count.

Agents can read this file before triggering a restart:

```bash
count=$(cat .claude/claudeclaw/active-runs 2>/dev/null || echo 0)
if [ "$count" -gt 1 ]; then
  # Post waiting notification via direct API call (NOT response text —
  # the session ends with a restart so buffered output is lost)
  curl -s -X POST "https://discord.com/api/v10/channels/$CHANNEL_ID/messages" \
    -H "Authorization: Bot $DISCORD_TOKEN" \
    -H "Content-Type: application/json" \
    -d "{\"content\": \"Waiting for $((count - 1)) other session(s) to finish...\"}"

  while [ "$(cat .claude/claudeclaw/active-runs 2>/dev/null || echo 0)" -gt 1 ]; do
    sleep 5
  done
fi
```

**Known limitation:** the check happens at LLM execution time (~30–60s after message receipt). Sessions shorter than that window may finish before the agent reads the file. This approach is intended for long-running tasks (minutes), not quick ones.

## Changes

- `src/runner.ts` — `mainRunCount` counter, `persistRunCount()` (sync write), `getMainRunCount()` export; increment/decrement wraps `execClaude` in try/finally

## Notes

The waiting notification must be sent via a direct transport API call (Bash tool `curl`), not as response text. ClaudeClaw delivers the session response only after the session completes — but this session ends with a restart, so buffered text is lost. See `RESTART_RESUME_SPEC.md` (PR #191) for the full wait-for-idle pattern.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.28",
+      "version": "1.0.29",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.28",
+  "version": "1.0.29",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -1,7 +1,7 @@
 import { mkdir, readFile, writeFile, realpath } from "fs/promises";
 import { join, dirname, resolve, sep } from "path";
 import { execSync } from "child_process";
-import { existsSync } from "fs";
+import { existsSync, writeFileSync, mkdirSync } from "fs";
 import {
   getSession,
   createSession,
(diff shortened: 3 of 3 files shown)


TerrysPOV:
Vision: ALIGNED — directly plugs the missing `active-runs` file gap flagged in the #191 review.

Complexity: SIMPLE — one file touched, minimal patch.

Overlapping PRs: #185 touches `runner.ts`; #82 touches `runner.ts`.

**Note: please rebase onto current master before merge — a PR was just merged.**

### Code review

Found 1 issue:

**`persistRunCount()` is called before `mkdir` — first-run write is silently dropped**

`mainRunCount++` and `persistRunCount()` fire at line 1001–1002, but `await mkdir(LOGS_DIR, { recursive: true })` is at line 1004. On a fresh install where `.claude/claudeclaw/` doesn't exist yet, the `writeFileSync` in `persistRunCount()` throws and is swallowed silently. The increment is never written — so during the first ever session, the file either doesn't exist or shows `0` instead of `1`. The decrement write at session end succeeds because `mkdir` has run by then. Fix: move `await mkdir(LOGS_DIR, { recursive: true })` before `mainRunCount++`, or add `mkdirSync(dirname(ACTIVE_RUNS_FILE), { recursive: true })` inside `persistRunCount()`.

https://github.com/moazbuilds/claudeclaw/blob/fdb900457f5b9a1e1ccae0d7ab368ffee1472927/src/runner.ts#L999-L1006

TerrysPOV (approved):
Rereviewed the latest branch tip.

The previous first-run `active-runs` directory issue is fixed by creating the target directory inside `persistRunCount()` before writing. I also checked this against current `master`; there are no rebase conflicts or version-bump issues to resolve.

Local validation passed:
- `git diff --check`
- merge-tree check against `master`
- `bun install`
- `bun test`
- `bunx tsc --noEmit --pretty false`

The remaining failed `claude-review` check is due to the known CI authentication issue (`ANTHROPIC_API_KEY` / 401), not this PR.

Clear to merge. Thanks for the contribution.

TerrysPOV:
Closing this as the work has been cherry-picked into #196, with the original author attribution preserved in the combined PR. Thanks @squirblej for the contribution.

PR #194, opened 2026-05-05
feat(discord): forward message support with coalescing

BCDel89:
## Summary

- Discord forwarded messages arrive with an empty `content` field — the text lives in `message_snapshots[0].message.content`. The early-exit guard was silently dropping them before the detection code ran. Fixed by adding a `hasForwardedContent` check.
- Discord delivers a forward + accompanying text comment as two separate MESSAGE_CREATE events, causing the bot to respond twice. Fixed with a 1.5s coalescing window: a pure forward is held in memory; if a text message from the same user+channel arrives within the window, the forward snapshot is absorbed as context into that message and the two are processed as one prompt.

## Test plan

- [ ] Forward a message with a text comment — one response that includes the forwarded content
- [ ] Forward a message with no comment — responds after ~1.5s with the forwarded content
- [ ] Normal messages (text, images, voice) unaffected
- [ ] Empty messages with no content/attachments/forward still filtered out

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.29",
+      "version": "1.0.33",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.32",
+      "version": "1.0.28",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.28",
+      "version": "1.0.29",
       "keywords": [
         "cron",
         "heartbeat",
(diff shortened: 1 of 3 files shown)


TerrysPOV:
Vision: ALIGNED — cleaner handling of Discord forwarded messages is a reasonable quality-of-life feature.

Complexity: SIMPLE — one file touched.

Overlapping PRs: #191 touches `discord.ts`; #185 touches `discord.ts`.

**Note: please rebase onto current master before merge — a PR was just merged.**

### Code review

Found 1 issue:

**Timer callback re-enters the deferral branch — the forward is deferred forever and never processed**

When the 1.5s timeout fires with no follow-up message, the callback deletes `pendingForwards[forwardKey]` then calls `handleMessageCreate(token, message)` with the original unmodified message. But that message still has `message_reference.type === 1`, no content, no attachments — so `isForwardOnly && hasForwardedContent` is true again. The now-empty `pendingForwards` map means `existing` is undefined, so a new timer is set and the function returns early again. This repeats every 1.5 seconds indefinitely; the forward is never passed to `runUserMessage`.

Fix: don't pass the original message back recursively. Instead, extract a `deliverForward(token, message, snapshot)` helper that skips the coalescing logic and delivers directly, and call that from the timer callback.

https://github.com/moazbuilds/claudeclaw/blob/5b514745cab44920aa59f2a39667edb16bbaba0f/src/commands/discord.ts#L637-L649

TerrysPOV (approved):
Approved. I reviewed the updated head (`a8ae960`) against current `master`.

The previously raised coalescing issue has been fixed: deferred forward delivery now bypasses the coalescing branch, so a pure forwarded message no longer loops indefinitely. I also added a small maintainer commit to correct the stale plugin/marketplace metadata bump to `1.0.33`.

Validation: merge-tree and `git diff --check` are clean, `bun test` passes, `bunx tsc --noEmit --pretty false` passes, and both plugin/marketplace version guards pass. The remaining `claude-review` failure is the known auth issue, not a PR-specific failure.

PR #200, opened 2026-05-07
feat(telegram): explicit model selection commands

Nate-br:
## Summary

Replaces the `/fast` toggle with explicit per-chat model selection commands, giving users clear control over which Claude model handles their messages.

**New commands:**
- `/model` — Show current model and available options
- `/modelhaiku` — Switch to Haiku (fastest, least capable)
- `/modelsonnet` — Switch to Sonnet (balanced, default)
- `/modelopus` — Switch to Opus (most capable, slower)
- `/modeldefault` — Reset to the model configured in settings

**Implementation:**
- Tracks model overrides per chat ID using `Map<number, string>`
- Passes the override to `runUserMessage()` as a new optional `modelOverride` parameter (threads through to `run()`)
- Removes the old `/fast` toggle (replaced by `/modelhaiku` + `/modeldefault`)
- Reorganises the command menu with emoji prefixes and logical groupings (session management / model selection / mode toggles / control)

**Why this change?**
The `/fast` toggle was ambiguous — users couldn't tell whether fast mode was on or off, and it only offered two states (default vs Haiku). Explicit model commands make the current model visible and support all three Claude model tiers.

## Validation

- [x] I ran the relevant checks locally
- [x] I updated any docs or setup guidance affected by this change

## Plugin Versioning

Plugin version bumped (`.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json`) as this change ships new Telegram commands under `src/`.

### DIFF
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -368,6 +368,12 @@ async function sendDocumentToChat(
 // Chat IDs with verbose tool display enabled
 const verboseChats = new Set<number>();
 
+// Model overrides per chat ID
+const chatModels = new Map<number, string>();
+const MODEL_HAIKU = "claude-haiku-4-5-20251001";
+const MODEL_SONNET = "claude-sonnet-4-6";
+const MODEL_OPUS = "claude-opus-4-7";
+
 /**
  * Build a streaming callback using editMessageText.
  * On first chunk: send a placeholder message to get message_id.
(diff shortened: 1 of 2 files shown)


TerrysPOV:
Thanks for the contribution — the model selection approach here is solid and `/fast`'s ambiguity was a real pain point worth fixing.

I've written up a proper PR description above based on your email reply, and rebased the branch onto current master (a bulk merge landed since you opened this).

One note for future PRs: a structured description is required before a PR can be reviewed or merged. The template asks for Summary, Validation, and Plugin Versioning — please fill it in when you open the PR rather than waiting to be asked. PRs opened without a description will be held until one is provided. It's a small thing but it makes a real difference when reviewing across multiple contributors.

Thanks again — looking forward to seeing more from you.

TerrysPOV (approved):

PR #202, opened 2026-05-07
feat(discord): stream incremental response preview while Claude works

squirblej:
## Summary

- POST a ⏳ placeholder when the first tool call fires (not at the end, so the user sees activity immediately)
- PATCH the placeholder every 1500ms with an italic preview of accumulated tool events and response text
- DELETE the preview and POST the final response when Claude finishes

## Configuration

Add to `settings.json`:
```json
{
  "discord": {
    "streaming": true
  }
}
```

Defaults to `false` — existing installs are unaffected.

## Preview format

Preview text is wrapped in Discord italic (`_…_`) so it's visually distinct from final responses. Underscores in content (file paths, variable names) are escaped to prevent broken formatting. Long previews show the most recent content with a `[...]` prefix.

## Tool event display

Tool calls appear as `● tool_name(arg)` and results as `  ⎿  [tool_name] summary`. This shows progress during long-running tool sequences before any response text has arrived.

## Implementation

- `src/runner.ts` — added optional `onChunk` / `onToolEvent` callbacks to `runClaudeOnce`, `runClaudeStream`, and `runUserMessage`
- `src/commands/discord.ts` — `makeDiscordStreamCallback()` manages the placeholder lifecycle; integrated into the message handler
- `src/config.ts` — `discord.streaming` boolean setting

## Test plan

- [ ] Send a message that triggers tool use — confirm ⏳ appears quickly and preview updates during tool execution
- [ ] Confirm final response replaces preview (not appended)
- [ ] Confirm `streaming: false` (default) shows no placeholder
- [ ] Confirm messages with no tool use still work (streaming only kicks in on first tool event)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.34",
+      "version": "1.0.35",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.34",
+  "version": "1.0.35",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -655,30 +655,25 @@ function makeDiscordStreamCallback(token: string, channelId: string): DiscordStr
     if (streamMsgId) scheduleEdit();
   };
 
-  const finalize = async (): Promise<void> => {
-    // Cancel any pending edit
-    if (editTimer) {
-      clearTimeout(editTimer);
-      editTimer = null;
-    }
-    // If placeholder was never posted, nothing to delete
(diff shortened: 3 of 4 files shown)


TerrysPOV (changes requested):
Thanks for this. The shape is good, but I think there is one blocking lifecycle bug before this is merge-ready.

With `discord.streaming` enabled, responses that do not emit any tool events can hang forever. The placeholder is only posted from `onToolEvent`, but after `runUserMessage` returns the Discord handler always does:

```ts
await streamCb.waitForStreamMsg();
await streamCb.finalize();
```

For a text-only Claude response, `onChunk` may run but `onToolEvent` never does, so `waitForStreamMsg()` never resolves and the final Discord message is never sent.

Please make the stream finalisation path complete even when no placeholder was started, e.g. by having a single completion/finalise method that resolves/nulls the waiter when no tool event occurred, or by avoiding the wait unless a placeholder post was actually started.

I checked the branch against current `master`: merge-tree and `git diff --check` are clean, `bun test` passes, and `bunx tsc --noEmit --pretty false` passes. The `claude-review` CI failure appears to be the known authentication issue rather than a PR failure.

TerrysPOV:
Closing this in favour of the bulk cherry-pick PR #209. Thank you @squirblej for the Discord streaming contribution; the original commits have been preserved in #209 so attribution is retained.

PR #203, opened 2026-05-07
fix(discord): restore thread delivery on gateway reconnects

squirblej:
## Problem

After a gateway reconnect, threads that the bot has active sessions in can go deaf — `MESSAGE_CREATE` events stop arriving for those threads. This happened reliably after `INVALID_SESSION` reconnects and sometimes after `RESUMED` reconnects.

Three root causes identified and fixed.

## Fix 1 — INVALID_SESSION: close WebSocket before re-IDENTIFY

When a non-resumable `INVALID_SESSION` arrives, the previous code sent `IDENTIFY` on the same WebSocket connection that just failed a `RESUME`. Discord appears not to fully restore thread `MESSAGE_CREATE` delivery for sessions established this way.

**Fix:** close the WebSocket (`ws.close(4000)`) and let the `onclose` handler open a fresh connection for `IDENTIFY`. The 4000 close code signals a clean client-initiated disconnect without triggering Discord's reconnect-loop detection.

## Fix 2 — RESUMED: skip REST rejoin entirely

The previous code did `DELETE + PUT` on all thread members even on a successful `RESUMED` reconnect, where the gateway session is still intact and delivery is already live. This was unnecessary at best and could disrupt subscriptions at worst.

**Fix:** `rejoinThreads` now takes a `trigger` parameter. On `RESUMED`, it returns immediately after logging (no REST calls).

## Fix 3 — GUILD_CREATE: PUT-only for already-joined threads

The `GUILD_CREATE` payload includes a `threads` array with a `member` field indicating whether the bot is already a subscribed member. For threads where `member` is set, a `DELETE` before the `PUT` is unnecessary and can disrupt the gateway subscription.

**Fix:** collect member thread IDs from `GUILD_CREATE`, skip the `DELETE` for those threads, `PUT` only.

## Diagnostic logging

`RECONNECT` and `INVALID_SESSION` opcode events are promoted from `debugLog` to `console.log` so reconnect cycles are always visible in production logs. `rejoinThreads` logs trigger type, session ID, member status, and PUT response codes.

## Test plan

- [ ] Trigger a `RECONNECT` (e.g. bot restart or network drop) — confirm `GUILD_CREATE` rejoin runs and threads remain responsive
- [ ] Trigger an `INVALID_SESSION` — confirm fresh WebSocket is opened and thread delivery is restored in the new session
- [ ] Confirm `RESUMED` reconnects skip REST calls and threads stay live
- [ ] Confirm threads not in `listenChannels` are still rejoined (they have sessions, just aren't listen-channels)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.34",
+      "version": "1.0.35",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.34",
+  "version": "1.0.35",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -400,18 +400,14 @@ async function rejoinThreads(
         // Not in GUILD_CREATE member list — force full rejoin to reset gateway subscription
         await discordApi(token, "DELETE", `/channels/${ts.threadId}/thread-members/@me`).catch(() => {});
       }
-      const putRes = await fetch(`${DISCORD_API}/channels/${ts.threadId}/thread-members/@me`, {
-        method: "PUT",
-        headers: { Authorization: `Bot ${token}` },
-      });
+      await discordApi(token, "PUT", `/channels/${ts.threadId}/thread-members/@me`);
       console.log(
-        `[Discord][REJOIN] thread=${ts.threadId} GUILD_CREATE=${isMember ? "member" : "non-member"} PUT=${putRes.status}`,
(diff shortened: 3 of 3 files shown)


TerrysPOV (changes requested):
Thanks for this, the reconnect direction makes sense.

I found one issue before this is merge-ready:

`rejoinThreads()` now performs the `PUT /channels/{thread}/thread-members/@me` call with raw `fetch()`, logs the status, and then always logs `Rejoined thread` regardless of whether Discord returned success. That means `403`, `404`, or `429` responses are treated as successful rejoins, and the code also bypasses the existing `discordApi()` rate-limit retry/error handling. In the exact reconnect path this PR is hardening, that can leave threads still unsubscribed while the logs say they were restored.

Please either use `discordApi(token, "PUT", ...)` again, or explicitly check `putRes.ok` and throw/log as a failed rejoin, including handling `429` consistently.

Local checks passed: `bun install`, `bun test`, `bunx tsc --noEmit --pretty false`, and diff/merge checks. The failing `claude-review` check is the known CI auth issue.

TerrysPOV:
Closing this in favour of the bulk cherry-pick PR #209. Thank you @squirblej for the Discord thread recovery contribution; the original commits have been preserved in #209 so attribution is retained.

PR #213, opened 2026-05-12
fix(slack): extract text from blocks/attachments when event.text is empty

### LINKED ISSUE #211: bug(slack): bot messages using Slack attachments are silently dropped due to empty text field
## Bug

When a Slack bot posts a message using the legacy **attachments** format (or block kit with no plain-text fallback), `event.text` is an empty string. The message handler in `slack.ts` bails out early:

```ts
if (!cleanText.trim() && !hasImage && !hasVoice && !hasDoc) return;
```

This silently drops the event — ClaudeClaw never responds, and no error is logged. The bug affects all bot messages that omit `text`, regardless of whether `allowBots` is configured.

## How to reproduce

Configure a Slack app (e.g. [Gatus](https://github.com/TwiN/gatus), Grafana, etc.) to post alerts into a channel ClaudeClaw listens to. These tools use Slack's attachment format:

```json
{
  "type": "message",
  "subtype": "bot_message",
  "text": "",
  "attachments": [
    {
      "title": "🚨 Alert",
      "text": "api.example.com is down.",
      "fields": [{ "title": "Condition results", "value": ":x: STATUS == 200" }]
    }
  ]
}
```

`event.text` is `""` → `cleanText.trim()` is `""` → message dropped.

## Fix

Fall back to extracting text from `attachments` (and `blocks`) when `event.text` is empty:

```ts
// Add to SlackMessage interface:
interface SlackAttachment {
  text?: string;
  fallback?: string;
  pretext?: string;
  title?: string;
  fields?: Array<{ title: string; value: string }>;
}
// SlackMessage:
  attachments?: SlackAttachment[];
  blocks?: SlackBlock[];

// In handleMessage, replace:
//   let cleanText = event.text;
// with:
let cleanText = event.text
  || (event.blocks?.length ? extractBlockText(event.blocks) : "")
  || (event.attachments?.length
    ? event.attachments.map(a => [
        a.pretext,
        a.title,
        a.text,
        ...(a.fields?.map(f => `${f.title}:\n${f.value}`) ?? []),
      ].filter(Boolean).join("\n")).join("\n")
    : "");
```

Where `extractBlockText` walks `blocks[].text.text`, `blocks[].fields[].text`, and `blocks[].elements[].text.text`.

## Notes

- Also related: bot messages often arrive with `subtype: "bot_message"`, which is currently filtered by `if (event.subtype && event.subtype !== "file_share") return`. If `allowBots` is in use (see #210), this subtype should also be allowed through. Suggested fix:

```ts
if (event.subtype && event.subtype !== "file_share" && !(isBotAllowed && event.subtype === "bot_message")) return;
```

- Discovered while implementing the `allowBots` feature from #210 and testing with Gatus alerts.

gwku:
Fixes #211.

## Problem

When a Slack bot posts using the legacy **attachments** format (Gatus, Grafana, PagerDuty, etc.), `event.text` is `""`. The handler falls through to:

```ts
if (!cleanText.trim() && !hasImage && !hasVoice && !hasDoc) return;
```

…and silently drops the event with no error logged. The same issue affects Block Kit messages without a plain-text fallback.

## Fix

Fall back through a chain of text sources before discarding the message:

1. `event.text` — existing behaviour, unchanged
2. Block Kit `blocks[]` — walks `block.text.text`, `block.fields[].text`, and `block.elements[].text.text`
3. Legacy `attachments[]` — joins `pretext`, `title`, `text`, and `fields[].title`/`value`

Adds `SlackAttachment` and `SlackBlock` TypeScript interfaces, `blocks?`/`attachments?` fields to `SlackMessage`, and an `extractBlockText()` helper.

## Test plan

- [ ] Message with `text: ""` + `attachments[]` to a listened channel — confirm response (previously silent drop)
- [ ] Block Kit message with no plain-text fallback — confirm content extracted from blocks
- [ ] Normal text messages unchanged
- [ ] Messages with no text, blocks, or attachments still dropped (existing behaviour)

Verified against real Gatus DOWN/RESOLVED alerts using the Slack attachments format.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.35",
+      "version": "1.0.36",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.35",
+  "version": "1.0.36",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/slack.ts
+++ b/src/commands/slack.ts
@@ -921,6 +921,7 @@ async function handleMessage(event: SlackMessage): Promise<void> {
   if (botUserId) {
     cleanText = cleanText.replace(new RegExp(`<@${botUserId}>`, "g"), "").trim();
   }
+  cleanText = sanitizeUserInput(cleanText);
 
   const files = event.files ?? [];
   const imageFiles = files.filter(isImageFile);
--- a/src/commands/slack.ts
+++ b/src/commands/slack.ts
@@ -58,6 +58,21 @@ interface SlackFile {
   filetype?: string;
 }
 
+interface SlackAttachment {
+  text?: string;
+  fallback?: string;
+  pretext?: string;
+  title?: string;
+  fields?: Array<{ title: string; value: string }>;
+}
+
+interface SlackBlock {
+  type: string;
+  text?: { type: string; text: string };
+  fields?: Array<{ type: string; text: string }>;
+  elements?: Array<{ type: string; text?: { type: string; text: string } }>;
+}
+
 interface SlackMessage {
   type: string;
   subtype?: string;
@@ -68,6 +83,8 @@ interface SlackMessage {
   ts: string;
   thread_ts?: string;     // set on thread replies; equals ts for the first reply
   files?: SlackFile[];
+  blocks?: SlackBlock[];
+  attachments?: SlackAttachment[];
   channel_type?: string;  // "im" | "mpim" | "channel" | "group"
 }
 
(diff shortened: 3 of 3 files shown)


TerrysPOV (changes requested):
Thanks Gerwin, this is a clean fix and the use case is bang on. Sorry it took a while for me to get to it.

I went ahead and bumped the plugin + marketplace versions in a maintainer commit so the version-guard CI checks would clear. Going forward could you run these yourself before opening a PR?

```
bun run bump:plugin-version
```

There's a note about it in CLAUDE.md — I should have pointed that out when we first chatted about this fix.

One thing I'd like you to address before I approve:

**Sanitize the fallback text.** The new `cleanText` (from `extractBlockText(blocks)` and the attachments join) lands in the Claude prompt at `promptParts.push(\`Message: ${cleanText}\`)` without going through `sanitizeUserInput()`. We added that to the thread-history path in `c335730` specifically to strip directive-like patterns from external content, and Gatus / Grafana / PagerDuty alert text is the same threat surface — webhook payloads often carry user-supplied check names that an attacker can influence. Could you wrap `cleanText` with `sanitizeUserInput()` once the fallback resolves? Should be a one-liner.

Heads-up: this overlaps with your other PR #214 — both touch the top of `handleMessage()`, so whichever of these lands second will hit a merge conflict. Not a blocker, just something to plan for.

Vision: aligned. Complexity: simple.

Once that sanitize wrap is in I'll approve and get it merged.

gwku (commented):
Good catch — added `sanitizeUserInput()` after the fallback resolves. Same line as the thread history path.

gwku (commented):
Good catch, added `sanitizeUserInput()` after the fallback resolves. Same line as the thread history path.

TerrysPOV (approved):
Sanitize wrap is exactly right, thanks Gerwin. Approving.

gwku (commented):
Thanks for the approval.

PR #214, opened 2026-05-12
feat(slack): add allowBots option for per-channel bot message pass-through

### LINKED ISSUE #210: feat(slack): per-channel allowBots option to allow bot messages to trigger responses
Currently `slack.ts` unconditionally drops all events where `event.bot_id` is set (line 838):

```ts
if (event.bot_id || !event.user) return;
```

This makes it impossible to use ClaudeClaw in channels that are exclusively or primarily written to by bots — monitoring channels being the most common example (e.g. Gatus, Grafana, PagerDuty alerting into a `#monitoring-alerts` channel).

**Proposed solution**

Add an optional `allowBots` array to the Slack config that lists channel IDs where bot messages should be passed through:

```json
"slack": {
  "botToken": "xoxb-...",
  "appToken": "xapp-...",
  "listenChannels": ["C123456"],
  "allowBots": ["C123456"]
}
```

The filter in `slack.ts` would become:

```ts
const allowBots = config.allowBots ?? [];
if (event.bot_id) {
  if (event.user === botUserId) return; // never respond to ourselves
  if (!allowBots.includes(event.channel)) return;
}
if (!event.bot_id && !event.user) return;
```

Note the self-exclusion check (`event.user === botUserId`) is critical — without it, ClaudeClaw will respond to its own bot messages in the thread and loop indefinitely.

**Context**

OpenClaw had this as a per-channel `allowBots: true` setting. Migrating from OpenClaw to ClaudeClaw breaks any workflow that relied on reacting to bot-posted alerts.

gwku:
Closes #210.

## Problem

The bot filter in `handleMessage` unconditionally drops all events where `event.bot_id` is set:

```ts
if (event.bot_id || !event.user) return;
```

This makes it impossible to react to messages from monitoring tools (Gatus, Grafana, PagerDuty, etc.) in dedicated alert channels, even when that is the intended use case.

## Solution

Add an optional `allowBots` array to the Slack config listing channel IDs where bot-posted messages should pass through:

```json
"slack": {
  "botToken": "xoxb-...",
  "appToken": "xapp-...",
  "listenChannels": ["C123456"],
  "allowBots": ["C123456"]
}
```

The guard becomes allowBots-aware:

```ts
const isBotAllowed = !!event.bot_id && allowBots.includes(event.channel);

if (event.bot_id) {
  if (event.user === botUserId) return; // never respond to our own messages
  if (!isBotAllowed) return;
}
```

The self-exclusion check (`event.user === botUserId`) prevents ClaudeClaw from responding to its own messages and looping indefinitely.

Bot messages in `allowBots` channels also bypass user-level auth (no human `userId` to check against `allowedUserIds`) and allow the `bot_message` subtype through alongside the existing `file_share` exception.

## Changes

- `src/config.ts`: add `allowBots: string[]` to `SlackConfig`, default, and parser
- `src/commands/slack.ts`: replace blanket bot guard with `allowBots`-aware logic

## Test plan

- [ ] Bot message in a channel listed in `allowBots` triggers a response
- [ ] Bot message in a channel not in `allowBots` is still dropped
- [ ] ClaudeClaw does not respond to its own messages (no loop)
- [ ] Normal user messages in all channel types unchanged
- [ ] `allowBots` omitted from config defaults to `[]` (existing behaviour preserved)

Verified locally against real Gatus DOWN/RESOLVED alerts.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.35",
+      "version": "1.0.36",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.36",
+      "version": "1.0.37",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.35",
+  "version": "1.0.36",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.36",
+  "version": "1.0.37",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/slack.ts
+++ b/src/commands/slack.ts
@@ -1066,7 +1066,7 @@ async function handleMessage(event: SlackMessage): Promise<void> {
     }, 20_000);
 
     const agentName = sessionThreadId
-      ? (() => { try { return agentDirKey(`slack-${channelId}`, event.thread_ts!); } catch { return undefined; } })()
+      ? (() => { try { return agentDirKey(`slack-${channelId}`, replyThreadTs); } catch { return undefined; } })()
       : undefined;
 
     let result;
--- a/src/commands/slack.ts
+++ b/src/commands/slack.ts
@@ -64,6 +64,8 @@ interface SlackMessage {
   text: string;
   user?: string;
   bot_id?: string;
+  username?: string;
+  bot_profile?: { name?: string };
   channel: string;
   ts: string;
   thread_ts?: string;     // set on thread replies; equals ts for the first reply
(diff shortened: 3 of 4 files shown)


TerrysPOV (changes requested):
Thanks Gerwin. Same as #213 — bumped the plugin + marketplace versions for you in a maintainer commit so the version-guard CI checks would clear. Going forward could you run these before opening a PR?

```
bun run bump:plugin-version
```

It's documented in CLAUDE.md — sorry, should have flagged this on both PRs sooner.

This one needs a bit more before I'm comfortable merging. Three things:

**1. Bot message text bypasses `sanitizeUserInput()`.** For an `allowBots` channel, `cleanText` flows straight into the Claude prompt with only the `@botUserId` mention stripped. The hardening in `c335730` was specifically about neutralising directive-like patterns (`[upload_file:...]`, `[read_channel:...]`, `[react:...]` etc.) coming from external content. A monitoring bot or a webhook with attacker-influenceable payload text is exactly the same surface. Could you wrap `cleanText` with `sanitizeUserInput()` on the bot-allowed path?

**2. No per-bot allowlist to replace the bypassed user auth.** The PR description rightly calls out that bot messages bypass `allowedUserIds`. The trouble is channel-scoped trust alone is pretty wide — any bot or webhook that can post in that channel can now drive the daemon. Could you add an optional `allowBotIds: string[]` (Slack `B...` app IDs) alongside `allowBots`, and require *both* channel and bot ID to match before the bypass kicks in? Leave it falling back to channel-only when empty if you want to preserve current behaviour, but I think we want it tight by default.

**3. `userId = event.user ?? event.bot_id` collapses bots into one session.** `bot_id` is the Slack application ID — same across every message from that app. So two different monitoring tools posting into the same `allowBots` channel will share an `agentName` / session context. Worth either keying off `event.username` or `event.bot_profile?.name` when present, or just adding a comment that this is intentional if you've thought about it and decided collapsing is fine.

Heads-up: overlaps with #213 — both touch the top of `handleMessage()`, so one of these will hit a merge conflict on landing.

Vision: partial — bounded by config but it does expand the attack surface. Complexity: simple by line count, but the auth bypass is meaningful.

Happy to do another pass once those land.

gwku (commented):
Addressed all three points:

1. `sanitizeUserInput()` now wraps `cleanText` on the bot-allowed path
2. Added `allowBotIds: string[]` alongside `allowBots` — when non-empty, both channel and bot ID must match. Empty defaults to channel-only (backwards compatible)
3. Non-thread bot messages now get an isolated session per `event.ts` so two bots in the same channel don't bleed into each other's context. `userId` label prefers `bot_profile.name` over `bot_id` for readable logs

TerrysPOV (approved):
All three points addressed cleanly, thanks.

I pushed one tiny maintainer commit (`fc08145`) on top — the `agentName` ternary was passing `event.thread_ts!` even on the non-thread bot path (where `thread_ts` is undefined), so `agentDirKey` was throwing and falling through to `undefined`. Net effect: `sessionThreadId` was correctly isolating context per bot, but the agent working directory was collapsing to the default across bots. Swapped it for `replyThreadTs` (which is `event.thread_ts ?? event.ts` — already in scope and matches the same ts that built `sessionThreadId`). Now non-thread bot messages get their own workdir as intended.

Bumped versions to 1.0.37 alongside.

Approving.

gwku (commented):
Good catch on the agentDirKey, missed that. Thanks for fixing it up.

PR #220, opened 2026-05-18
fix(slack): cancel proactive reconnect timer on close and stop

### LINKED ISSUE #219: bug(slack): proactive reconnect timer leaks on each openSocket call, cascades into rapid reconnect loop and process death
## Bug

Every call to `openSocket` schedules a 29-minute proactive reconnect via `setTimeout`, but the return value is never stored and therefore can never be cancelled. When the socket closes and reconnects, the old timer is still running. After a few reconnect cycles the timers stack up and start firing within seconds of each other — each firing closes the socket, which triggers another reconnect, which spawns another timer. The process eventually exhausts itself and dies.

## How to reproduce

Let the Slack bot run for ~1 hour with a few natural reconnects (Socket Mode rotates every 30 min). Watch `journalctl -u claudeclaw-slack` — after the second or third rotation, "Proactive reconnect (30-min rotation)" starts firing every few seconds. The service dies shortly after.

## Root cause

In `openSocket` (`src/commands/slack.ts`):

```ts
// timer reference is discarded — cannot be cancelled
setTimeout(() => {
  if (!running) return;
  ws?.close(1000, "Proactive reconnect");
}, RECONNECT_MS);
```

`reconnectTimer` (for the post-close reconnect delay) is handled correctly with a module-level reference — the proactive timer just missed the same treatment.

## Fix

Add a module-level `proactiveReconnectTimer`, cancel it before setting a new one in `openSocket`, clear it in `ws.onclose`, and clear it in `stopSlack`.

```ts
let proactiveReconnectTimer: ReturnType<typeof setTimeout> | null = null;

// in openSocket, replace the bare setTimeout:
if (proactiveReconnectTimer) clearTimeout(proactiveReconnectTimer);
proactiveReconnectTimer = setTimeout(() => {
  proactiveReconnectTimer = null;
  if (!running) return;
  ws?.close(1000, "Proactive reconnect");
}, RECONNECT_MS);

// in ws.onclose, before scheduleReconnect:
if (proactiveReconnectTimer) { clearTimeout(proactiveReconnectTimer); proactiveReconnectTimer = null; }

// in stopSlack, alongside the reconnectTimer cleanup:
if (proactiveReconnectTimer) { clearTimeout(proactiveReconnectTimer); proactiveReconnectTimer = null; }
```

Will open a PR.

gwku:
Fixes #219.

## Problem

`openSocket` schedules a 29-minute proactive reconnect timer but discards the return value of `setTimeout`, so the timer can never be cancelled. On each reconnect (Socket Mode rotates every ~30 min) a new timer is created while old ones keep running. After a few cycles they stack up and start firing seconds apart — each one closes the socket, triggers another reconnect, and spawns yet another timer. The process eventually dies.

## Fix

Store the timer in a module-level `proactiveReconnectTimer` variable, matching the existing pattern for `reconnectTimer`:

- Cancel any existing timer before setting a new one in `openSocket`
- Clear it in `ws.onclose` (socket closed before the timer fired)
- Clear it in `stopSlack` (shutdown cleanup)

Same three-line pattern already used for `reconnectTimer` — the proactive one just missed the same treatment.

## Test plan

- [ ] Let the bot run through two or more 30-minute Socket Mode rotations and confirm "Proactive reconnect" fires once per cycle, not repeatedly
- [ ] Confirm the service stays alive across multiple reconnects
- [ ] `stopSlack()` / restart clears the timer cleanly with no dangling callbacks

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.1.1",
+      "version": "1.0.38",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.1.1",
+  "version": "1.0.38",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/slack.ts
+++ b/src/commands/slack.ts
@@ -116,6 +116,7 @@ let ws: WebSocket | null = null;
 let running = true;
 let slackDebug = false;
 let reconnectTimer: ReturnType<typeof setTimeout> | null = null;
+let proactiveReconnectTimer: ReturnType<typeof setTimeout> | null = null;
 
 // Dedup: track recently processed message timestamps to avoid handling both message + app_mention
 const recentlyProcessed = new Map<string, number>();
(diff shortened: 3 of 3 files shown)


TerrysPOV (approved):
Thanks Gerwin. Clean fix — mirrors the existing `reconnectTimer` cancel pattern from `923a293` exactly. All four locations from #219 are covered (module-level decl, cancel-before-set in `openSocket`, clear in `ws.onclose`, clear in `stopSlack`).

I rebased the branch onto master so #185's v1.1.0 security changes go in first, then resolved the version-file conflict and bumped to 1.1.2 in a maintainer touch-up. No other conflicts — slack.ts isn't touched by #185.

Approving.

PR #221, opened 2026-05-18
fix(bus): wire real Discord gateway + REST so production can flip to bus runtime

TerrysPOV:
## Summary

- Port the legacy `src/commands/discord.ts` WS gateway loop (lines 1582–1908) into a class-shaped `DiscordGateway` in `src/adapters/discord/gateway.ts` and the REST subset (sendMessage chunking + interaction callback) into `src/adapters/discord/rest-api.ts`.
- `DiscordAdapter` now auto-constructs both from its token when no `gateway`/`restApi` is injected. Tests keep injecting `FakeDiscordGateway` / `FakeDiscordRestApi` so the suite stays offline.
- Adds focused unit tests for both new modules (HELLO/IDENTIFY/RESUME handshake, MESSAGE_CREATE + INTERACTION_CREATE dispatch, RECONNECT close behaviour; REST chunking, `[react:…]` stripping, interaction callback shape, non-OK propagation).

## Why

Last night's Hetzner cutover crashed instantly because the Bus `DiscordAdapter` was constructed with `bus + token + routing` (no `gateway` injection), and the Sprint 3 constructor threw:

> DiscordAdapter: \`gateway\` injection is required in Sprint 3. Shared gateway helper extraction is a Sprint 4 deliverable; until then production callers must construct their own gateway.

That Sprint 4 follow-up never landed. This PR is the smallest viable port: no shared helper extraction across both runtimes (legacy `src/commands/discord.ts` is untouched), just enough to make `runtime: "bus"` mount the Discord adapter without a crash.

## What is NOT in this PR

The legacy listener also handles GUILD_CREATE / THREAD_* (thread rejoin), slash-command registration, attachment downloads + voice transcription, and streaming edit-in-place rendering. Those are orthogonal to gateway construction and can land incrementally; the staging guide (PR #128) already calls them out as future work.

## Test plan

- [x] \`bun test src/adapters/discord\` — 38 pass / 0 fail (was 1 fail before the test for "rejects construction with no gateway" was rewritten as "auto-constructs when none injected")
- [x] \`bun test src/bus src/adapters\` — 328 pass / 0 fail / 1 skip
- [x] \`bun run lint\` — 0 errors (baseline 938 warnings unchanged)
- [x] \`bun run bump:plugin-version\` + \`bun run bump:marketplace-version\` → 2.2.60
- [ ] Re-attempt Hetzner cutover to \`runtime: "bus"\` once merged


TerrysPOV:
Wrong repo — this work belongs on the Plus fork (TerrysPOV/ClaudeClaw-Plus). Re-opening there. Sorry for the noise.

PR #222, opened 2026-05-19
feat: auto-compact resumed sessions when turn count crosses threshold

royvergara:
## The problem

ClaudeClaw resumes the same Claude session across every Telegram (or Discord/Slack) message it receives, which is the right behavior for continuity — but the runner currently only warns once at `COMPACT_WARN_THRESHOLD` (25 turns) and otherwise lets sessions grow forever.

On my deployment, the daemon's Telegram session sat at **677 turns over 20 days**. Every new "What's the error?" prompt was paying the input-token cost for all 677 prior turns, and the account's rolling-window cap started returning `Credit balance is too low` on every single message. Meanwhile a fresh, interactive Claude Code session on the same account worked perfectly — exactly because its context wasn't 677 turns deep. The user never sees this in their Claude usage dashboard because per-account limits aggregate; they just see the daemon stop responding.

The existing `compactWarned: true` in `session.json` confirms the runner had already noticed and warned, but no consumer wired up `onCompactEvent` to surface the warning, and there was no automatic mitigation.

## The fix

Add an automatic `/compact` triggered after a successful turn once `turnCount` crosses a configurable threshold (default **100**):

- **`AutoCompactConfig` in `Settings`** (`autoCompact: { enabled, threshold }`). Defaults to `{ enabled: true, threshold: 100 }`; existing `settings.json` files keep working because the parser falls back to defaults when fields are missing.
- **`resetTurnTracking()` / `resetThreadTurnTracking()`** helpers in `sessions.ts` and `sessionManager.ts` that zero `turnCount` and clear the one-shot `compactWarned` flag after a successful compact, so the threshold doesn't immediately re-fire on the next turn.
- **Two new `CompactEvent` variants** (`auto-compact-by-turns-start`, `auto-compact-by-turns-done`) so any surface can notify the user — they're distinct from the existing timeout-driven `auto-compact-{start,done}` events.
- **Telegram listener** in `commands/telegram.ts` that DMs the first allowed user on `warn` + `auto-compact-by-turns-{start,done}` events. (Discord and Slack can hook up the same pattern in a follow-up — I didn't touch those because I don't run them and didn't want to ship untested code.)
- **Plugin + marketplace version bumped** `1.0.37 → 1.0.38`.

The timeout-driven auto-compact path (`COMPACT_TIMEOUT_ENABLED`, exit 124) is untouched.

## How to verify

```bash
# In settings.json, set a low threshold so it fires quickly:
"autoCompact": { "enabled": true, "threshold": 5 }
```

Send 5-6 Telegram messages. After the 5th successful turn:
- `daemon.log` shows `Auto-compact: turn 5 >= threshold 5`
- Telegram DM arrives: "🧹 Auto-compacting session…" then "✅ Auto-compact complete."
- `.claude/claudeclaw/session.json` shows `turnCount: 0` and `compactWarned: false`
- Same `sessionId` — the Claude session continues, just with a compacted transcript

To disable entirely:
```json
"autoCompact": { "enabled": false, "threshold": 100 }
```

## What I didn't do (would love feedback)

- I kept the default at 100 conservatively. Real cost-vs-continuity sweet spot probably varies by usage pattern.
- Discord + Slack don't subscribe to the new events yet — happy to add in this PR if you'd like, or a follow-up.
- Considered adding a `lastCompactedAt` field but the existing `compactWarned` flag plus `turnCount` reset gives us enough state for now.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.37",
+      "version": "1.0.38",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.37",
+  "version": "1.0.38",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -1,4 +1,4 @@
-import { ensureProjectClaudeMd, run, runUserMessage, runFork, killActive, isMainBusy, compactCurrentSession, compactCurrentThreadSession, isRateLimited, getRateLimitResetAt, getPermissionMode, setPermissionMode, type PermissionMode } from "../runner";
+import { ensureProjectClaudeMd, run, runUserMessage, runFork, killActive, isMainBusy, compactCurrentSession, compactCurrentThreadSession, isRateLimited, getRateLimitResetAt, getPermissionMode, setPermissionMode, onCompactEvent, type PermissionMode } from "../runner";
 import { extractErrorDetail } from "../messaging";
 import { loadPendingResume } from "../pending-resume";
 import { getSettings, loadSettings } from "../config";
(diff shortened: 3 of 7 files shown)


royvergara:
Heads up — the `claude-review` job on this PR is failing with `Failed to authenticate. API Error: 401 Invalid authentication credentials` ([run](https://github.com/moazbuilds/claudeclaw/actions/runs/26103014781/job/76759006294)). Looks like `secrets.CLAUDE_CODE_OAUTH_TOKEN` in this repo has expired or been rotated and the workflow's copy hasn't been refreshed. Worth checking other recent PRs to see if they're hitting the same — if so, a one-time secret refresh in repo Settings → Secrets and variables → Actions should clear it. Not blocking this PR; just flagging.

TerrysPOV (changes requested):
Thanks Roy, the auto-compact feature is well-motivated — the long-running Telegram sessions accumulating unbounded transcript cost has been on my radar for a while.

I'd like one blocker addressed before I approve, plus a few secondary asks. Also a heads-up on the rebase: PR #185 just merged, which changes the version target and touches two of the same files this PR does.

## Blocker — auto-compact loops on failure

In `src/runner.ts`, when `runCompact` returns `false`, `resetTurnTracking` is **not** called. `turnCount` stays at or above `autoCompact.threshold`, so the very next successful turn re-enters the `turnCount >= threshold` branch and fires another compact, fails again, emits another `❌ Auto-compact failed` DM, and so on. Every turn after the first failure becomes a compact attempt + failure notification until something finally succeeds.

Fix shape — either:
- Reset `turnCount` to 0 even on failure (simple, accepts the risk of skipping retries), or
- Add a `compactAttemptedAt` timestamp or `compactFailCooldown` counter in session state so the threshold doesn't re-trigger for N turns after a failed attempt.

## Secondary asks

1. **No tests.** Auto-compact destroys transcript context — safety-critical new behaviour with zero coverage. Worth at minimum: threshold-crossing fires compact, success resets counter, failure path doesn't loop (the blocker above).

2. **Synchronous compact blocks the turn reply.** `runCompact` is awaited inline inside `execClaude` after the user message has been processed. The next user-visible reply is delayed by the full compact duration with no per-call timeout. Either document that this is intentional or move to a background-with-promise pattern.

3. **`config.ts` doc comment is misleading.** The comment says "threshold ≤ `COMPACT_WARN_THRESHOLD` (25) effectively disables the trigger", but the actual guard is `autoCompact.threshold > 0` — any value 1-24 still fires auto-compact, silently, before the warn threshold. Either enforce `threshold >= COMPACT_WARN_THRESHOLD` in `parseSettings` or fix the comment.

4. **Edge case — `threshold == COMPACT_WARN_THRESHOLD` double-notifies.** When set to exactly 25, both the warn event and the auto-compact-by-turns-start event fire on the same turn. Worth either preventing the warn when auto-compact will trigger on the same turn, or documenting that the two are independent.

5. **Discord/Slack notifier gap.** The PR explicitly defers this. The `onCompactEvent` bus is already in place so adding the equivalents should be a small follow-up — happy for it to be a separate PR if you'd rather not bloat this one, but worth flagging that Slack/Discord users will be silently compacted until then.

## Rebase notes — PR #185 has merged

Master now has the v1.1.0 security hardening changes. When you rebase:

- **Version bump target is 1.1.1** (master is currently at 1.1.1 after my maintainer follow-up on #185, so you'll need to bump to 1.1.2). Re-run `bun run bump:plugin-version` + `bun run bump:marketplace-version` after rebase.

- **`src/commands/telegram.ts` line 1** — hard conflict. Master now has both `import { wrapUntrusted } from "../prompt-safety"` and `import { isAllowed } from "../allowlist"`. Your change adds `onCompactEvent` to the existing `runner` import. Resolve by keeping all three imports — they're separate statements.

- **`src/config.ts`** — soft conflict on `DEFAULT_SETTINGS`. Master now has `allowedGuilds` in the discord block; your `autoCompact` addition is at a different key. Git will likely flag the context window — just accept both.

- **`src/runner.ts`** — no textual conflict expected. #185's untrusted-content hunks are ~280 lines above your auto-compact block.

Happy to do another pass once those land. Thanks for the work.

PR #225, opened 2026-05-27
ui: auto-build dist/web on startWebUi when missing

NorthIsUp:
## Problem

When clawdcode is installed via `claude plugin install` (vs. running from a checkout), the source lands in `~/.claude/plugins/cache/<marketplace>/<plugin>/<version>/` without a `dist/web/`. The build is a dev-time `bun run build:web` and isn't carried in the plugin tarball.

Result: `/ui/`, `/darwin/`, `/os9/`, `/osish/` all 404 with "UI not built — run \`bun run build:web\`" until a human kubectl-execs in and runs it.

## Fix

Detect the missing `dist/web/ui/index.html` sentinel at `startWebUi()` entry and run `bun run build:web` synchronously before `Bun.serve` binds. `node_modules/` is already present after `claude plugin install` (claude-cli runs `bun install` as part of install), so the tailwind binary the build relies on is on disk.

## Verified

In a live install (cluster-deployed v1.0.76 instance):
- Before: `/ui/` returns the "UI not built" 404.
- Manual `bun run build:web` in the plugin cache dir builds 4 bundles in <1s.
- With this PR: same thing happens automatically on first daemon start.

## Trade-offs

- First start after a fresh install is ~1s slower (one synchronous build). Subsequent starts no-op because the sentinel exists.
- If the build fails, the daemon still starts and serves the API — only the UI routes 404. Failure is logged to stderr.


NorthIsUp:
wrong target — meant for NorthIsUp/clawdcode fork

PR #233, opened 2026-06-04
fix(commands): resolve /context JSONL path without daemon cwd (#229)

rudi193-cmd:
Fixes #229. Adds findSessionJsonlPath in src/sessionFiles.ts; wires telegram/discord /context and ui sessions/usage. Tests: tests/session-files.test.ts (5 pass).

PR #234, opened 2026-06-04
fix(sessions): treat session.json without sessionId as absent (#228)

### LINKED ISSUE #228: Crash loop: TypeError when session.json exists but sessionId is missing
## Bug

`runner.ts` crashes in a tight loop when `session.json` exists but is missing the `sessionId` field. The daemon restarts via watchdog, crashes again on the next heartbeat, and repeats indefinitely.

## Root cause

`getSession()` in `sessions.ts` reads `session.json` and returns the parsed object. If the file contains `{"turnCount":0,"compactWarned":false,"lastUsedAt":"..."}` (no `sessionId` or `createdAt`), `getSession()` returns a **truthy** object with `sessionId: undefined`.

In `runner.ts` (~line 1010):

```ts
const isNew = !existing;
```

`existing` is truthy (it's an object), so `isNew` is `false`. Then ~line 1047:

```ts
console.log(
  `... (${isNew ? "new session" : `resume ${existing!.sessionId.slice(0, 8)}`} ...)`
);
```

`existing!.sessionId` is `undefined`, so `.slice()` throws:

```
TypeError: undefined is not an object (evaluating 'existing.sessionId.slice')
```

## How the corruption happens

I'm not 100% sure what originally wrote the incomplete `session.json`, but it may be a race between a session write and a crash/kill, or an edge case in the session-creation path where the file is written before `sessionId` is assigned.

## Fix

```diff
- const isNew = !existing;
+ const isNew = !existing || !existing.sessionId;
```

This treats a corrupted session (missing `sessionId`) as a new session, triggering a fresh bootstrap instead of crashing.

A secondary hardening could also be applied at line ~1047:

```diff
- `resume ${existing!.sessionId.slice(0, 8)}`
+ `resume ${existing!.sessionId?.slice(0, 8) ?? "unknown"}`
```

## Environment

- ClaudeClaw v1.0.27
- Bun runtime on Windows 10
- Crash confirmed in daemon log with 3+ restart cycles before manual intervention

## Workaround

Delete the corrupted `session.json` and restart the daemon. A fresh session bootstraps normally.

rudi193-cmd:
Fixes #228. Corrupted session.json without sessionId is treated as absent; runner bootstraps instead of crashing on sessionId.slice. Tests: tests/sessions-missing-id.test.ts

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.39",
+      "version": "1.0.40",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.40",
+      "version": "1.0.41",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.39",
+  "version": "1.0.40",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.40",
+  "version": "1.0.41",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -1050,7 +1050,7 @@ async function execClaude(
   const existing = threadId
     ? await getThreadSession(threadId)
     : await getSession(agentName);
-  const isNew = !existing;
+  const isNew = !existing?.sessionId;
   // Start the watchdog clock for resumed sessions (we know the ID immediately).
   if (existing) startSession(existing.sessionId);
   const timestamp = new Date().toISOString().replace(/[:.]/g, "-");
--- a/src/sessionManager.ts
+++ b/src/sessionManager.ts
@@ -1,4 +1,5 @@
 import { join } from "path";
+import { hasValidSessionId } from "./sessionValidate";
 
 const HEARTBEAT_DIR = join(process.cwd(), ".claude", "claudeclaw");
 const SESSIONS_FILE = join(HEARTBEAT_DIR, "sessions.json");
@@ -40,7 +41,7 @@ export async function getThreadSession(
 ): Promise<{ sessionId: string; turnCount: number; compactWarned: boolean } | null> {
   const data = await loadSessions();
   const session = data.threads[threadId];
-  if (!session) return null;
+  if (!hasValidSessionId(session)) return null;
 
   if (typeof session.turnCount !== "number") session.turnCount = 0;
   if (typeof session.compactWarned !== "boolean") session.compactWarned = false;
--- a/src/sessionManager.ts
+++ b/src/sessionManager.ts
@@ -107,5 +107,6 @@ export async function listThreadSessions(): Promise<ThreadSession[]> {
 /** Peek at a thread session without updating lastUsedAt. */
 export async function peekThreadSession(threadId: string): Promise<ThreadSession | null> {
   const data = await loadSessions();
-  return data.threads[threadId] ?? null;
+  const session = data.threads[threadId];
+  return hasValidSessionId(session) ? session : null;
 }
--- a/src/sessionManager.ts
+++ b/src/sessionManager.ts
@@ -98,10 +98,10 @@ export async function markThreadCompactWarned(threadId: string): Promise<void> {
   await saveSessions(data);
 }
 
-/** List all active thread sessions. */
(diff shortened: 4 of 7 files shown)


rudi193-cmd:
Note on the failing `claude-review` check: this is the same workflow credential failure as #233, before review runs:

```text
ANTHROPIC_API_KEY:
Failed to authenticate. API Error: 401 Invalid authentication credentials
```

The patch verification passed locally before push:

```text
node --experimental-strip-types --test tests/sessions-missing-id.test.ts
# 3 pass, 0 fail
```

So this PR is blocked on the repo's Claude Code Review workflow credentials / rerun, not on a code failure from this change.

TerrysPOV (changes requested):
### Code review

Found 1 issue:

1. `peekThreadSession` is the one remaining read path that doesn't go through `hasValidSessionId`. A corrupted `sessions.json` thread row — exactly the corruption class issue #228 reports — passes the `if (!session)` check in the Telegram/Discord `/status` and `/context` handlers, then crashes on `session.sessionId.slice(0, 8)` (or the `${session.sessionId}.jsonl` interpolation), reproducing the same `Cannot read properties of undefined (reading 'slice')` this PR is meant to eliminate.

`peekThreadSession` returns the raw record without validation:

https://github.com/moazbuilds/claudeclaw/blob/404ba91b0c1b19411382558a17cd945336520be3/src/sessionManager.ts#L106-L112

Vulnerable callers:

https://github.com/moazbuilds/claudeclaw/blob/404ba91b0c1b19411382558a17cd945336520be3/src/commands/telegram.ts#L1073-L1085

https://github.com/moazbuilds/claudeclaw/blob/404ba91b0c1b19411382558a17cd945336520be3/src/commands/discord.ts#L1183-L1196

Quickest fix: route `peekThreadSession` through the same `hasValidSessionId` guard `getThreadSession` now uses (lines 41-45 of `sessionManager.ts`), so it returns `null` for corrupted rows.

rudi193-cmd:
Addressed the review feedback in 5a6e633.

`peekThreadSession` now routes through the same `hasValidSessionId` guard as `getThreadSession`, so corrupted `sessions.json` thread rows return `null` instead of reaching callers that call `session.sessionId.slice(...)`.

Added a regression test in `tests/sessions-missing-id.test.ts`.

```sh
node --experimental-strip-types --test tests/sessions-missing-id.test.ts
# 4 pass, 0 fail
```

Ready for another look when you have a moment.

rudi193-cmd:
@TerrysPOV friendly re-review ping — `peekThreadSession` now routes through the same `hasValidSessionId` guard as `getThreadSession` (`5a6e633`), so corrupted `sessions.json` thread rows return `null`. `claude-review` is green and the branch is mergeable. Re-requested your review; happy to adjust if anything else stands out.

TerrysPOV (changes requested):
### Code review

`peekThreadSession` blocker is resolved in 5a6e6336 — thanks.

Found 1 remaining issue (same class):

1. `listThreadSessions` is the last unfiltered thread read path. A corrupted `sessions.json` row will be returned as-is, and the `/status` thread-sessions loop in Discord crashes on `ts.sessionId.slice(0, 8)` — same `TypeError` signature this PR is meant to eliminate.

`listThreadSessions` returns the raw map values without `hasValidSessionId`:

https://github.com/moazbuilds/claudeclaw/blob/5a6e6336c3ddb9f61fd722af5b816fdd59aaa48f/src/sessionManager.ts#L100-L105

Crash site:

https://github.com/moazbuilds/claudeclaw/blob/5a6e6336c3ddb9f61fd722af5b816fdd59aaa48f/src/commands/discord.ts#L1169-L1175

Quickest fix, mirroring the `peekThreadSession` shape in 5a6e6336:

```ts
return Object.values(data.threads).filter(hasValidSessionId);
```

(Sister mutators `incrementThreadTurn` / `markThreadCompactWarned` only bump fields and bail on `!session`, so they don't crash — they just mutate corrupted rows. Worth a follow-up but not blocking.)

rudi193-cmd:
Resolved in a9eb7cb. `listThreadSessions` now filters through `hasValidSessionId`, so a corrupted `sessions.json` row can no longer reach the `/status` thread-sessions loop and crash on `ts.sessionId.slice(0, 8)`. That was the last unfiltered thread read path — `getThreadSession`, `peekThreadSession`, and `listThreadSessions` are now all gated. Thanks for the catch.

TerrysPOV:
@claude

TerrysPOV (approved):
The `listThreadSessions` blocker from the prior review is resolved — `Object.values(data.threads).filter(hasValidSessionId)` in a9eb7cb. I traced the full crash class and it's now complete: every path that dereferences `.sessionId` (Discord/Telegram/Slack /status + /context, runner, rotation, web UI) is sourced from a validated read, and the web-UI readers stay safe via their parallel `UUID_RE.test()` guard.

Maintainer-committed the version bump correction in c5e6343: the branch's 1.0.40 bump had become a no-op after #233 landed master at 1.0.40, so I re-bumped to 1.0.41 to clear plugin-version-guard / marketplace-version-guard. Please run `bun run bump:plugin-version` + `bun run bump:marketplace-version` yourself on future PRs — and rebase if another version bump lands before this merges.

Approving.

TerrysPOV (approved):
Approved

TerrysPOV (approved):
Re-approving on the current head 54854a9 after reverting my erroneous 1.0.41 bump.

Context: the code fix is complete — the listThreadSessions guard (a9eb7cb) closes the last unvalidated thread-session read path, and I traced the full crash class to confirm no unguarded .sessionId dereference remains. The author's 1.0.40 bump already satisfies plugin-version-guard / marketplace-version-guard (they compare against this PR's base at 1.0.39, and both ran green at this SHA). My earlier extra bump to 1.0.41 was a misdiagnosis that introduced a merge conflict against master's 1.0.40; I've force-pushed it out, so the branch is back to a clean, mergeable state.

Approving.

PR #235, opened 2026-06-04
Fix for discord threads

### LINKED ISSUE #230: Discord thread handling issue / OpenClaw makes stuff on main channel
If I create a thread on discord in a #channel, which ClaudeClaw is configured to listen to, following happens
- ClaudeClaw picks up conversation in this thread and handles the prompt in separate session (correct)
- on main #channel, ClaudeClaw sees stripped info about this channel first message and tries to handle it on the main session (wrong). Moreover the channel only contains some part of thread first message, and ClaudeClaw treats this as a prompt, in many cases just asking for more info.

There seems to be no way to disable thread creation recap on the channel, nor reconfigure the bot. I would think ClaudeClaw should rather ignore new thread creation info on the parent #channel.

azaz44:
## Summary

ClaudeClaw will now ignore system messages (like thread creation) and only listen to messages and replies.

Fixes: #230 

## Validation

- [x] I ran the relevant checks locally
- [ ] I updated any docs or setup guidance affected by this change - not needed I believe

## Plugin Versioning

Versions bumped.

Run as needed:

```bash
bun run bump:plugin-version
```

Typical rule:
- bump `.claude-plugin/plugin.json` when shipped plugin content changes
- bump `.claude-plugin/marketplace.json` when marketplace metadata should reflect the new shipped version

Docs-only and other non-shipped changes do not require these bumps.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.39",
+      "version": "1.0.40",
       "keywords": [
         "cron",
         "heartbeat",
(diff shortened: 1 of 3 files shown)


TerrysPOV:
Closing in favour of #239

PR #237, opened 2026-06-06
Multi-Session Threads For Telegram

tamircher:
Each Telegram topic already gets its own session, and the runner keeps independent per-thread serial queues (`threadQueues`) — but the Telegram handler gates every incoming message on the global `isMainBusy()`, so any in-flight run anywhere blocks **all** topics.

This PR adds per-thread busy tracking in the runner (`isThreadBusy`/`isGlobalBusy`) and checks only the target topic's queue (or the global session for non-topic chats). Each topic becomes a fully independent session running in parallel; a second message to the *same* topic still gets the polite busy reply.

**Changes**
- `src/runner.ts`: per-thread busy set + global counter, maintained in `execClaude`'s entry/finally alongside `mainRunCount`
- `src/commands/telegram.ts`: busy check scoped to the message's `sessionKey` instead of global

### DIFF
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -1,4 +1,4 @@
-import { ensureProjectClaudeMd, run, runUserMessage, runFork, killActive, isMainBusy, compactCurrentSession, compactCurrentThreadSession, isRateLimited, getRateLimitResetAt, getPermissionMode, setPermissionMode, type PermissionMode } from "../runner";
+import { ensureProjectClaudeMd, run, runUserMessage, runFork, killActive, isThreadBusy, isGlobalBusy, compactCurrentSession, compactCurrentThreadSession, isRateLimited, getRateLimitResetAt, getPermissionMode, setPermissionMode, type PermissionMode } from "../runner";
 import { wrapUntrusted } from "../prompt-safety";
 import { isAllowed } from "../allowlist";
 import { extractErrorDetail } from "../messaging";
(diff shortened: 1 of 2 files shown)


TerrysPOV (changes requested):
### Code review

Found 1 issue:

1. `/kill` becomes a cross-topic footgun once parallel topics are enabled. `killActive()` iterates the global `mainActiveProcs` set and SIGTERMs every entry — before this PR that set could only ever hold one proc, so kill-all was kill-mine. After this PR, a `/kill` issued from topic A also terminates the in-flight Claude run in topics B, C, etc. Codex flagged this exact concern on PR #184 ("Track active subprocesses per task before exposing /kill") — the Set was added then, but `killActive` was never made thread-aware.

`killActive` is global:

https://github.com/moazbuilds/claudeclaw/blob/fc2b9f6772c63bc9ca60f49e1e83fdef5414f44c/src/runner.ts#L283-L295

Telegram `/kill` handler invokes it without scoping to the caller's topic:

https://github.com/moazbuilds/claudeclaw/blob/fc2b9f6772c63bc9ca60f49e1e83fdef5414f44c/src/commands/telegram.ts#L1148-L1153

Quickest fix: track procs per-thread (e.g. `Map<string | "__global__", Set<Proc>>` keyed on `threadId` from the `execClaude` call site) and have `killActive(threadId)` only iterate its own bucket. Telegram handler passes the message's `sessionKey`.

Also missing the mandatory plugin + marketplace version bumps (`.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json`). I can maintainer-commit those once the `killActive` fix lands.

PR #238, opened 2026-06-06
feat(telegram): queue messages sent mid-run instead of rejecting them

tamircher:
**Stacked on #237 — review the last commit only; I'll rebase once #237 lands.**

A message arriving while the same topic (or the global session) is mid-run is rejected with a busy notice, forcing the user to resend. Since `run()` already serializes per thread via `enqueue()`, the message can simply be submitted — it executes next, in order, with full session context via `--resume`, like Claude Code's message queue.

While queued, the bot reacts 👀 to the incoming message as a lightweight received-and-waiting ack. The typing indicator already runs while queued, and the stream preview message is only created on the first output chunk, so nothing else changes visibly until the queued run starts.

**Changes**
- `src/commands/telegram.ts`: drop the busy rejection; submit to the per-thread queue and 👀-react while waiting

### DIFF
--- a/src/commands/telegram.ts
+++ b/src/commands/telegram.ts
@@ -1444,24 +1444,22 @@ async function handleMessage(message: TelegramMessage): Promise<void> {
       );
     }
     const prefixedPrompt = promptParts.join("\n");
-    // Per-thread busy check: only reject if THIS topic's queue (or the global
-    // session for non-topic chats) is mid-run. Different topics run in parallel.
+    // Per-thread queue: if THIS topic (or the global session for non-topic
+    // chats) is mid-run, the message queues behind the current run — like
+    // Claude Code — instead of being rejected. run() serializes per thread,
+    // so queued messages execute in order with full session context. React
+    // 👀 so the user knows the message was received and is waiting its turn.
     const busy = sessionKey ? isThreadBusy(sessionKey) : isGlobalBusy();
+    if (busy) {
+      await sendReaction(config.token, chatId, message.message_id, "👀").catch(() => {});
+    }
     const verbose = verboseChats.has(chatId);
     const modelOverride = chatModels.get(chatId);
-    let result;
-    let streamMsgId: number | null = null;
-    let hadToolLines = false;
-    if (busy) {
-      await sendMessage(config.token, chatId, "Claude is still working on the previous message in this topic — try again in a moment, or use /fork for a quick parallel task.", threadId);
-      return;
-    } else {
-      const stream = makeStreamCallback(config.token, chatId, threadId, { verbose });
-      result = await runUserMessage("telegram", prefixedPrompt, sessionKey, undefined, stream.onChunk, stream.onToolEvent, modelOverride);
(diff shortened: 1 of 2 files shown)


TerrysPOV (changes requested):
### Code review

Nice swap from reject to queue — session context is correctly preserved through `enqueue → execClaude → --resume` (the 0f8a351 rationale for the rejection was the auto-fork path, not queueing, so this isn't a regression).

Two correctness concerns worth flagging before merge, plus one note:

1. **Stale clock prefix on queued messages.** `runUserMessage` builds the `[YYYY-MM-DD HH:MM:SS UTC±X]` prefix synchronously at call time, then enqueues the closure. A message that sits behind a 5-minute run is handed to Claude with a 5-minute-stale "now". Time-aware prompts ("what time is it?", "schedule in 30 min", "is it past my bedtime?") will silently get wrong answers. Pre-PR the reject-and-retry pattern always rebuilt a fresh prefix on the user's resend.

   https://github.com/moazbuilds/claudeclaw/blob/37c95e070a78a4713a27703f36bb4080b918efa7/src/runner.ts#L1697-L1700

   Fix: move `prefixUserMessageWithClock(prompt)` inside the enqueued closure so it stamps at execution time instead of arrival time.

2. **`/kill` no longer means "stop everything".** `killActive()` only SIGTERMs entries in `mainActiveProcs`, which are the currently-running subprocesses. Queued messages aren't in there yet — they're closures chained on `threadQueues`. So `/kill` after spamming a topic kills the in-flight run, then the queue happily fires each queued message one-by-one, spinning up a new subprocess for each. User intent is silently broken. This is distinct from the cross-topic kill issue on #237.

   https://github.com/moazbuilds/claudeclaw/blob/37c95e070a78a4713a27703f36bb4080b918efa7/src/runner.ts#L283-L295

   Fix: when `/kill` fires, also drain or replace the matching `threadQueues` entry with `Promise.resolve()` so pending closures are short-circuited.

Note (not blocking): the per-thread queue has no length cap. For a personal bot gated by `isAllowed` this is fine; worth a follow-up if you ever broaden access.

Also: the runner.ts comment block introduced in #237 still says "while still rejecting a second message to the SAME thread" — stale after this PR:

https://github.com/moazbuilds/claudeclaw/blob/37c95e070a78a4713a27703f36bb4080b918efa7/src/runner.ts#L299-L308

Once #237 lands and you rebase, please run `bun run bump:plugin-version` + `bun run bump:marketplace-version` (will need the next-after-237's bump).

TerrysPOV:
Closing in favour of #250, which implements the same fix — removing the busy-rejection so mid-run Telegram messages queue instead of being dropped — as a simpler standalone change that doesn't depend on the #237 stack. Consolidating the two competing PRs onto one.

Thanks for this, @tamircher — the approach was right, and #250 carries it forward. The per-thread parallel-topics work in #237 is a separate feature and stays open on its own track.

The same open items from the earlier review still apply and have been raised on #250 (queue backpressure, `/kill` not cancelling queued messages, stale clock prefix), so nothing is lost by consolidating here.

PR #239, opened 2026-06-12
fix(discord): ignore thread recap system messages (#230)

### LINKED ISSUE #230: Discord thread handling issue / OpenClaw makes stuff on main channel

rudi193-cmd:
## Summary

Fixes #230.

Discord emits system messages (e.g. type 18 thread-creation recap) on the parent channel when a thread is created. Those were routed through `handleMessageCreate` and could trigger spurious Claude prompts.

Only process regular messages (type 0) and replies (type 19). Bump plugin/marketplace version to 1.0.40 per contributor guidelines.

**Note:** #235 implements the same filter; this PR is rebased on current `master` from `rudi193-cmd` in case maintainers prefer a fresh branch. Happy to close in favor of #235 if that lands first.

## Test plan

- [ ] Create a Discord thread in a listened channel — parent channel should not get a Claude prompt from the recap message
- [ ] Normal messages and replies in threads still work

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.39",
+      "version": "1.0.40",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.39",
+  "version": "1.0.40",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/discord.ts
+++ b/src/commands/discord.ts
@@ -79,6 +79,12 @@ interface DiscordMessage {
   type: number;
 }
 
+const enum DiscordMessageType {
+  Default = 0,
+  Reply = 19,
+  ThreadCreated = 18,
+}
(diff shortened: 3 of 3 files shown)


rudi193-cmd:
Note on the failing `claude-review` check: this is failing before review runs because the action has no Claude credential available in the repo:

```text
ANTHROPIC_API_KEY:
Failed to authenticate. API Error: 401 Invalid authentication credentials.
```

The check shows as SUCCESS in the rollup because the workflow exits 0 on auth failure — but mergeStateStatus is UNSTABLE. This is a repo-side secret configuration issue, not a problem with the PR code. Happy to address any actual code feedback.

TerrysPOV (approved):
Reviewed — fix is clean. Enum + filter placement is correct (after bot-author check, before auth); no code concerns.

Heads-up: byte-identical to #235 (azaz44, draft) and the bumps to 1.0.40 are now stale since #233 landed master at 1.0.40 — version-guard will need 1.0.41 on rebase. Thanks for the parallel implementation @rudi193-cmd.

PR #242, opened 2026-06-30
feat(config): support configurable fallback base URLs

### LINKED ISSUE #215: feat: make fallback base URL configurable via settings.json
# Revised GitHub Issue: Configurable Fallback Base URL

### Problem
Currently, the fallback model's API base URL is hardcoded to `https://api.z.ai/api/anthropic` in `src/runner.ts`. This restricts users to using only GLM as a fallback and prevents the use of other Anthropic-compatible providers like **OpenRouter**, **LiteLLM**, or custom local proxies.

Additionally, the current logic is gated behind the hardcoded string `"glm"`. This means even if a user wants to use a different provider (like Gemini or DeepSeek via a proxy), they are currently forced to name their model "glm" in `settings.json` just to trigger the custom fallback address.

### Proposed Solution
Add an optional `baseUrl` field to the configuration and refactor the environment building logic to be **provider-agnostic**. If a `baseUrl` is provided in the configuration, it should be used for the `ANTHROPIC_BASE_URL` regardless of what the model is named.

---

#### 1. `src/config.ts`
Add an optional `baseUrl` property to the `ModelConfig` interface so both primary and fallback models can use custom endpoints:
```typescript
export interface ModelConfig {
  model: string;
  api: string;
  baseUrl?: string; // New optional field
}
```

#### 2. `src/runner.ts`
Update `buildChildEnv` to prioritize the user-defined `baseUrl`. This removes the rigid dependency on the "glm" string while maintaining backward compatibility:

```typescript
function buildChildEnv(baseEnv: Record<string, string>, config: ModelConfig): Record<string, string> {
  const childEnv: Record<string, string> = { ...baseEnv };
  const normalizedModel = config.model.trim().toLowerCase();

  if (config.api.trim()) childEnv.ANTHROPIC_AUTH_TOKEN = config.api.trim();

  // Priority 1: User-defined baseUrl from settings.json
  // Priority 2: Legacy hardcoded fallback for "glm" model
  if (config.baseUrl) {
    childEnv.ANTHROPIC_BASE_URL = config.baseUrl;
    childEnv.API_TIMEOUT_MS = "3000000";
  } else if (normalizedModel === "glm") {
    childEnv.ANTHROPIC_BASE_URL = "https://api.z.ai/api/anthropic";
    childEnv.API_TIMEOUT_MS = "3000000";
  }

  return childEnv;
}
```

#### 3. Execution Logic
Ensure that both the primary and fallback execution paths in `execClaude` and `runClaudeStream` pass their respective `ModelConfig` objects to the environment builder.

---

### Benefits
- **Flexibility**: Support for any Anthropic-compatible API provider (OpenRouter, DeepSeek, Gemini-via-proxy, etc.).
- **Improved DX**: No need to "trick" the system by naming models "glm" to get a custom URL.
- **Backward Compatibility**: Existing GLM configurations remain fully functional.

vibe-cy:
Problem
ClaudeClaw currently ties the fallback provider base URL to the special `glm` model name. This makes fallback routing provider-specific and forces users of Anthropic-compatible providers such as OpenRouter, LiteLLM, or local proxies to rely on a hardcoded GLM endpoint or misleading model names.

Solution
This PR adds explicit `baseUrl` support to the model configuration and routes the complete model config into Claude subprocess environment construction. Explicit base URLs now set `ANTHROPIC_BASE_URL`; the existing `model: "glm"` behavior remains intact as the legacy default when no base URL is configured.

Changes
- Add `baseUrl` / `fallback.baseUrl` parsing to settings.
- Refactor runner environment construction to accept full `ModelConfig` objects.
- Preserve legacy GLM endpoint behavior when `baseUrl` is omitted.
- Update config/start command docs for fallback base URL usage.
- Add focused tests for settings parsing and environment generation.
- Add a minimal demo script for inspecting fallback base URL env injection.
- Bump plugin and marketplace metadata to 1.0.41.

Testing
- `bun test src/__tests__/config.test.ts src/__tests__/runner-env.test.ts src/__tests__/jobs.test.ts`
  - 18 pass / 0 fail
- `bun run examples/fallback-base-url/demo.ts`
  - confirms `ANTHROPIC_AUTH_TOKEN` and `ANTHROPIC_BASE_URL` are populated from fallback config
- `bun test`
  - 127 pass / 3 fail
  - Remaining failures are pre-existing HOME-path expectations in `tests/session-files.test.ts` and match the baseline observed before this change.

Notes for Reviewer
- Explicit `baseUrl` takes precedence over the GLM default endpoint.
- `model: "glm"` without `baseUrl` still maps to `https://api.z.ai/api/anthropic` for backward compatibility.
- `sameModelConfig` now includes `baseUrl` so a fallback can differ from the primary by provider endpoint even when model/API fields overlap.
- I kept the change dependency-free and limited to the existing config/runner path.

Closes #215

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.44",
+      "version": "1.0.45",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.44",
+  "version": "1.0.45",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/commands/config.md
+++ b/commands/config.md
@@ -154,6 +154,16 @@ Set or clear the API token for the fallback model.
 4. Set `fallback.api` to the new value.
 5. Write and confirm.
 
+### `fallback base-url <url>` / `fallback base-url`
+
+Set or clear the Anthropic-compatible base URL for the fallback model.
+
+1. If URL is in `$ARGUMENTS`, use it directly.
+2. Otherwise, use **AskUserQuestion**: "What base URL should ClaudeClaw use for fallback model?" (header: "Fallback URL", options: let user type via Other)
+3. Read `.claude/claudeclaw/settings.json`.
+4. Set `fallback.baseUrl` to the new value (`""` to clear).
+5. Write and confirm.
+
 ### `timezone <tz>` / `timezone`
 
 Set the IANA timezone (e.g. `America/New_York`, `Europe/London`, `UTC`).
@@ -251,9 +261,11 @@ Location: `.claude/claudeclaw/settings.json`
 {
   "model": "opus",
   "api": "",
+  "baseUrl": "",
   "fallback": {
     "model": "glm",
-    "api": ""
+    "api": "",
+    "baseUrl": ""
   },
   "timezone": "America/New_York",
   "timezoneOffsetMinutes": -300,
@@ -286,9 +298,11 @@ Location: `.claude/claudeclaw/settings.json`
 | Key                        | Type       | Description                                    |
 |----------------------------|------------|------------------------------------------------|
 | `model`                    | string     | Claude model (`opus`, `sonnet`, `haiku`, `glm`, or full ID). Empty = default |
-| `api`                      | string     | API token used when model is `glm` (mapped to `ANTHROPIC_AUTH_TOKEN`) |
+| `api`                      | string     | API token mapped to `ANTHROPIC_AUTH_TOKEN` for the primary model |
+| `baseUrl`                  | string     | Optional Anthropic-compatible base URL for the primary model |
 | `fallback.model`           | string     | Backup model used automatically if primary run returns rate-limit text (recommend `glm` for provider diversity) |
 | `fallback.api`             | string     | API token used with `fallback.model` (optional) |
+| `fallback.baseUrl`         | string     | Optional Anthropic-compatible base URL for fallback providers |
 | `timezone`                 | string     | IANA timezone name (e.g. `America/New_York`)   |
 | `timezoneOffsetMinutes`    | number     | UTC offset in minutes (auto-resolved from timezone) |
 | `heartbeat.enabled`        | boolean    | Whether the recurring heartbeat runs           |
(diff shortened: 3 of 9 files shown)


TerrysPOV (changes requested):
[COMPLEX — triage] 9 files, but mitigated: +194/-46 (well under 300 lines), substance is in two code files (`config.ts`, `runner.ts`); the rest are tests, docs, one example, and version files. Clean refactor of the existing config/runner env path — no new process model or plugin API.

Vision: ALIGNED — strictly opt-in. With `baseUrl` unset (every existing install), the default path is byte-identical, and the previously-implicit GLM→z.ai endpoint becomes explicit and configurable rather than hidden. Docs match the code; the `buildChildEnv` export the example relies on is real.

Overlapping PRs: the `1.0.41` version bump collides with **#234, which merged earlier today and already put master at 1.0.41** — this branch (last updated 2026-06-30) needs a rebase onto current master + re-bump to **1.0.42**, or both version guards fail. Also shares `runner.ts` regions with open #222 / #251 (merge-time friction only, no logical overlap).

### Code review

Found 2 issues:

1. Ambient `ANTHROPIC_API_KEY` is forwarded to a user-specified `baseUrl`. `cleanSpawnEnv` copies the whole parent env and strips only the three `CLAUDE_CODE*` keys — not `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN`:

https://github.com/moazbuilds/claudeclaw/blob/829e7eceffa32670f624f1524d85f597dcc3895e/src/runner.ts#L107-L119

`buildChildEnv` sets `ANTHROPIC_AUTH_TOKEN` only when a config `api` is present and points `ANTHROPIC_BASE_URL` at the custom URL, but never clears an inherited Anthropic key:

https://github.com/moazbuilds/claudeclaw/blob/829e7eceffa32670f624f1524d85f597dcc3895e/src/runner.ts#L334-L354

So a user who has `ANTHROPIC_API_KEY` exported (a standard Claude Code auth method), points `baseUrl`/`fallback.baseUrl` at a third-party proxy, and leaves `api` blank ("just route through my proxy, no separate key") ships their real Anthropic key to that host. Pre-PR this couldn't happen for arbitrary hosts — only the hardcoded z.ai endpoint under `model: glm`. Suggest: when a custom `baseUrl` is set and no `api` is configured, drop `ANTHROPIC_API_KEY`/`ANTHROPIC_AUTH_TOKEN` from the child env (the same posture #102 established for host-managed tokens).

2. The 50-minute request timeout isn't extended for custom `baseUrl`s. `API_TIMEOUT_MS = "3000000"` is still gated on `model === "glm"` only:

https://github.com/moazbuilds/claudeclaw/blob/829e7eceffa32670f624f1524d85f597dcc3895e/src/runner.ts#L349-L351

The whole point of the feature is slower Anthropic-compatible providers (OpenRouter, LiteLLM, local proxies), but those get the default client timeout where glm doesn't — so "my OpenRouter fallback times out but glm didn't." Extend the timeout when a custom `baseUrl` is set, not just for glm.

PR #248, opened 2026-07-05
feat(discord): channel-scoped allowlist for guild messages

BCDel89:
## Summary

- Adds `discord.channelAllowedUserIds` config (`channelId -> user IDs[]`) — grants a user access to one specific guild channel without adding them to the global `allowedUserIds` list
- Channel-scoped access **never** applies to DMs, even for the same user ID — the global `allowedUserIds` list remains the only path to DM access
- Extracted the composed auth check into `isDiscordAuthorized()` in `allowlist.ts` so it's unit-testable independent of the Discord message-handling flow

## Why

Previously `allowedUserIds` was all-or-nothing: a user was either allowed to reach Sage everywhere (DMs + every guild channel), or blocked everywhere. No way to grant someone scoped access to a single project channel.

## Testing

- 7 new unit tests in `tests/allowlist.test.ts` covering: global allowlist access in DMs and guild channels, channel-scoped access in the correct channel, explicit denial in other channels, explicit denial via DM (the key leak this feature must avoid), denial for unlisted users, and no-throw on missing config
- Full suite: 117/117 passing, no regressions
- `tsc --noEmit`: no new errors introduced
- Manually verified live against Discord: a channel-scoped test user got real responses in their assigned channel, no response in other guild channels, and an explicit "Unauthorized." on DM — while the existing globally-allowed user was unaffected in all contexts

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.40",
+      "version": "1.0.41",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.42",
+      "version": "1.0.43",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.42",
+  "version": "1.0.43",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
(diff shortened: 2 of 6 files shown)


TerrysPOV (approved):
Well-built feature. I ran an adversarial fail-open/bypass audit on the auth logic and it's fail-closed on every path: `isDiscordAuthorized` checks the global allowlist first, gates the channel-scoped branch on `isGuild`, and ends with a terminal `return false`. Verified in particular:
- DM bypass is impossible — channel-scoped entries are `isGuild`-gated, so DMs (no `guild_id`) can only match the global allowlist. Locked by a test.
- Empty/missing/malformed-shape config fails closed (`?? []` → `isAllowed([])` → false).
- Wrong-channel access, type confusion, threads, and slash/button interactions are all deny-direction, and #185's front-door hardening is preserved (global-only interaction gate untouched).

Maintainer-committed the version housekeeping in 006a0d8: the branch's 1.0.41 bump conflicted with master (now 1.0.42 after #234/#247), so I merged current master in and re-bumped to 1.0.43. Please rebase on latest master and run the bump scripts yourself for future PRs.

Two non-blocking follow-ups:
1. The `config.ts` parser's sanitization of malformed `channelAllowedUserIds` (non-object, string-instead-of-array, numeric IDs) isn't unit-tested — the "malformed config can't grant access" property is only asserted for the `undefined` shape. The code is safe; a couple of `parseSettings` tests would pin it.
2. Two safe-but-undocumented asymmetries worth a doc line: channel-scoped users can send messages but not use slash-commands/buttons (interaction gate is global-only), and their access doesn't extend into threads (distinct channel_id).

Approving.

PR #255, opened 2026-07-17
fix(slack): make attachments visible in thread/channel history; pass all mention images

vincezh2000:
## Problem

`fetchThreadHistory` and `fetchChannelHistory` serialize only `msg.text` — the `files` field is never parsed. A message that is just a file upload (a screenshot, an invoice photo, a PDF) renders as an empty `[User Uxxx]:` line, so the model cannot even tell a file existed.

Real-world failure: a user posts several invoice images in a channel, then mentions the bot with "summarize the invoices above". The bot fetches history, sees blank lines, and answers "I don't see any invoice — please re-upload into this thread". The information was there the whole time.

A related gap on the triggering message itself: only `imageFiles[0]` is downloaded, so a mention carrying several images silently drops all but the first.

## Change

- **History attachments become visible.** Both history fetchers parse `files[]` and render one note per attachment:
  - downloaded → `[image saved: /path]` / `[file "name" saved: /path]` — reuses the existing `downloadSlackFile` (same inbox dir, same 25MB cap and extension allowlist), newest-first, capped at `HISTORY_FILE_DOWNLOAD_CAP = 6` per fetch to bound latency on low-spec machines;
  - voice → `[voice message — not transcribed]` (no transcription cost for history);
  - everything else (over cap, no `url_private`) → `[attachment "name" (type) — not downloaded]` — still visible, so the model can ask for it.
- **Read-your-attachments hint.** When a history block contains a saved path, it appends one instruction line telling the model to read the relevant files before answering. Channel history notes go through the existing `sanitizeUserInput`, same as message text.
- **Mentions pass up to 5 images** instead of only the first; the single-image prompt wording is unchanged, so existing behavior is identical for the 1-image case.

## Tests

`tests/slack-history.test.ts` covers the pure helpers (`formatHistoryFileNote`, `appendFileNotes`, the instruction line in `formatThreadHistoryAsContext`), including the headline case: a file-only message no longer renders as an empty line. Suite result is unchanged apart from the 9 new passing tests (the 3 pre-existing `sessionFiles` failures fail identically on master without this change).

Version bumps included per CONTRIBUTING (`bump:plugin-version` / `bump:marketplace-version` → 1.0.41).

## Deliberately out of scope

Rate-limit-aware caching/429 handling for history fetches: Slack's 2025 non-Marketplace terms throttle `conversations.history`/`replies` hard (~1 req/min, 15/page) for affected apps, which argues for a per-channel cache + `Retry-After` handling around these fetchers. We run that in production as a downstream patch, but it's opinionated enough that it felt like a separate discussion — happy to follow up with a second PR if there's interest.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "claudeclaw",
       "source": "./",
       "description": "Cron-like daemon that runs Claude prompts on a schedule",
-      "version": "1.0.40",
+      "version": "1.0.41",
       "keywords": [
         "cron",
         "heartbeat",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "claudeclaw",
-  "version": "1.0.40",
+  "version": "1.0.41",
   "description": "Cron-like daemon that runs Claude prompts on a schedule"
 }
--- a/src/commands/slack.ts
+++ b/src/commands/slack.ts
@@ -600,15 +600,18 @@ async function fetchThreadHistory(
       user?: string;
       bot_id?: string;
       ts: string;
+      files?: SlackFile[];
     }>;
   };
   if (!data.ok) {
     throw new Error(`conversations.replies error: ${data.error ?? "unknown"}`);
   }
 
-  return (data.messages ?? []).map((msg) => ({
+  const msgs = data.messages ?? [];
+  const resolved = await downloadHistoryFiles(token, msgs);
+  return msgs.map((msg) => ({
     role: msg.bot_id ? "assistant" : "user",
-    text: msg.text,
+    text: appendFileNotes(msg.text, msg.files, resolved),
     user: msg.user,
     ts: msg.ts,
   }));
@@ -625,6 +628,9 @@ function formatThreadHistoryAsContext(
     lines.push(`[${sender}]: ${msg.text}`);
   }
   lines.push("--- End of Thread History ---");
+  if (historyHasSavedAttachment(messages.map((m) => m.text))) {
+    lines.push("(Attachments in the history above are saved to local paths — read the relevant ones directly before answering.)");
+  }
   lines.push("");
   return lines.join("\n");
 }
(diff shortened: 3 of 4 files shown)


TerrysPOV (changes requested):
Genuinely useful fix — file-only Slack messages rendering as blank `[User]:` lines is a real bug, and the implementation is careful: path handling is safe (downloads reuse the hardened `downloadSlackFile`, keyed on Slack's `file.id` not the attacker-supplied `file.name`, with the extension allowlist and 25 MB cap intact — #185's hardening is preserved), newest-first cap logic is correct, text-only messages are unchanged, and you included real behavioural tests. The 5-image mention change is clean.

One thing needs addressing before this can ship, plus the version.

### Code review

1. The new history rendering feeds attacker-controllable content into the prompt unfenced, and instructs the model to read attacker-authored files directly. Both history helpers append `(Attachments … saved to local paths — read the relevant ones directly before answering.)` and inject the transcript with only plain `--- Thread History ---` delimiters:

https://github.com/moazbuilds/claudeclaw/blob/73243a2b961ef0539b3ab6c020f23883b5b43bdf/src/commands/slack.ts#L624-L633

`sanitizeUserInput` only strips ClaudeClaw's own `[bracket]` directives — it does not neutralize free-form injection prose, and it can't touch the contents of a downloaded screenshot/PDF. `slack.ts` never uses `wrapUntrusted`, so unlike discord.ts / telegram.ts this history text lands in the prompt outside the `<untrusted-…>` fence that the runner's anti-injection system prompt actually keys on. The feature auto-downloads up to 6 attachments from *any* participant across recent history and then tells the model to open them — so a member who uploads an image/PDF whose visible text says "ignore prior instructions and …" now gets that content read as part of answering an unrelated request. This widens the existing surface (Slack was already unfenced) rather than opening a new class, but it's a real widening — and notably it's *less* guarded than the sibling channel-read path already in this file, which does label its content:

https://github.com/moazbuilds/claudeclaw/blob/73243a2b961ef0539b3ab6c020f23883b5b43bdf/src/commands/slack.ts#L1303-L1303

Please wrap the rendered history + attachment notes in `wrapUntrusted(...)` the way discord/telegram wrap user content (that's the mechanism the runner's fence depends on), or at minimum carry the same `[… untrusted external content …]` labeling that line 1303 uses, and soften the bare "read the relevant ones directly before answering" imperative for multi-user historical attachments.

2. Version is stale — targets 1.0.41 but master is 1.0.43, which is the only thing making the PR conflict (slack.ts merges clean). Rebase and re-bump both `.claude-plugin/*.json` to 1.0.44.

Non-blocking:

- `HISTORY_FILE_DOWNLOAD_CAP` bounds *successful* downloads but not *attempts* — `downloaded++` only fires on success, so a history full of failing or oversized attachments will still attempt a fetch per file, each buffering up to 25 MB via `arrayBuffer()` before the post-read size check rejects. Bounded, but a latency/resource vector on the low-spec machines the project targets; consider capping attempts.
- `formatHistoryFileNote` does `(f.name ?? f.id).slice(0, 80)`, which throws if a file has neither field — inconsistent with the nearby `!f.id` guard. A `?? ""` fallback hardens it (Slack always sends `id`, so low severity).
- `historyHasSavedAttachment` matches the literal substring `" saved: "` anywhere in the rendered text, so a user message containing that string triggers the "read the relevant ones" advisory spuriously. Cosmetic.

